PR #18, opened 2026-05-10
Fix: ctypes 64-bit truncation on L40 (sm_89) + FP8 parameter support

wangerforcs:
## Environment

- GPU: NVIDIA L40 (Compute Capability 8.9, Ada Lovelace)
- Native Linux (no Docker)

## Problem 1: Segfault from ctypes 64-bit pointer/stream truncation

**Root cause**: ctypes defaults all function parameters to `c_int` (32-bit) when `argtypes` is not set. On 64-bit systems, CUDA runtime functions like `cudaMemcpyAsync` receive 64-bit device pointers and `cudaStream_t` handles. Without explicit `argtypes`, ctypes silently truncates these values to 32 bits, causing segfaults.

The stream handle value `97327966053360` is clearly a 64-bit pointer, but ctypes was passing it as a 32-bit `c_int` to the CUDA runtime.

**Fix** (in `cuda_buffer.py`):

```python
_cudart.cudaMemcpyAsync.argtypes = [ctypes.c_void_p, ctypes.c_void_p, ctypes.c_size_t, ctypes.c_int, ctypes.c_void_p]
_cudart.cudaMemcpyAsync.restype = ctypes.c_int
```

This tells ctypes that `src`/`dst` are 64-bit pointers (`c_void_p`) and `stream` is a 64-bit handle (`c_void_p`), preventing truncation.

Additionally, `pi05_rtx.py` was importing its own `ctypes.CDLL("libcudart.so")` instance instead of sharing the one from `cuda_buffer.py`. Now it imports `_cudart` from `cuda_buffer` to ensure the argtypes are applied consistently.

## Problem 2: FP8 not configurable

Previously `use_fp8` was hardcoded to `True` in the pipeline construction, with no way for users to disable it. On GPUs that don't fully support FP8 (e.g. sm_89), this causes issues.

**Fix**: Added a `use_fp8` parameter to `load_model()` (default `True`) and `Pi05TorchFrontendRtx.__init__()`. Users can now pass `use_fp8=False` to disable FP8 when needed.

```python
# Before: hardcoded
use_fp8=True, use_fp8_decoder=True

# After: configurable via parameter
use_fp8=self.use_fp8, use_fp8_decoder=self.use_fp8
```

## Files changed

- `flash_rt/core/cuda_buffer.py` — add argtypes for `cudaMemcpy` and `cudaMemcpyAsync`
- `flash_rt/api.py` — add `use_fp8` parameter to `load_model()`
- `flash_rt/frontends/torch/pi05_rtx.py` — import shared `_cudart` from `cuda_buffer`, add `use_fp8` parameter, use instance variable instead of hardcoded `True`
- `flash_rt/models/pi0/pipeline_rtx.py` — FP8 default adjustment

## Test
```python
import numpy as np
import flash_rt

print("[1/3] Generating random images ...")
base_img = np.random.randint(0, 256, (224, 224, 3), dtype=np.uint8)
wrist_img = np.random.randint(0, 256, (224, 224, 3), dtype=np.uint8)
print(f"  base_img: shape={base_img.shape}, dtype={base_img.dtype}")

print("[2/3] Loading model...")
model = flash_rt.load_model(
    checkpoint="pi05_pytorch_checkpoint", #using libero checkpoint
    config="pi05",
    framework="torch",
    use_fp8=False,
)

print("[3/3] Running predict...")
for i in range(10):
    actions = model.predict(images=[base_img, wrist_img], prompt="pick up the red block")
print(f"Done. actions shape: {actions.shape}")
```

output 
```shell
[1/3] Generating random images ...
  base_img: shape=(224, 224, 3), dtype=uint8
[2/3] Loading model...
[3/3] Running predict...
  autotune: tested 8 algos, best=0 (12.8864 us)
  autotune: tested 8 algos, best=0 (13.0048 us)
Done. actions shape: (10, 7)
```

### DIFF
--- a/flash_rt/api.py
+++ b/flash_rt/api.py
@@ -179,7 +179,8 @@ def load_model(checkpoint, framework="torch", num_views=2, autotune=3,
                fp4_layers=None,
                use_awq=None,
                awq_alpha=0.5,
-               use_p1_split_gu=None):
+               use_p1_split_gu=None,
+               use_fp8=True):
     """Load a FlashRT model.
 
     Args:
@@ -234,6 +235,9 @@ def load_model(checkpoint, framework="torch", num_views=2, autotune=3,
         fp4_layers: Tuple of encoder layer indices to FP4-quantize (only
             applies when use_fp4=True). Default (7, 8, 9) = middle FFN
             subset, LIBERO-validated. Other subsets untested at task level.
+        use_fp8: Enable FP8 execution where the selected frontend supports
+            an FP8/BF16 switch. Defaults to True to preserve existing
+            performance-oriented behavior.
 
     Returns:
         VLAModel instance with .predict() method.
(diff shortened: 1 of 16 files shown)


LiangSu8899:
Thanks a lot for the detailed investigation and fixes — really appreciate it!

PR #37, opened 2026-05-20
Add Pi0.5 Thor FP16 CUTLASS+megakernel encoder path

LiangSu8899:
## Summary
- Adds the SM110 (Thor) FP16 encoder GEMM path on top of the existing `use_fp8=False` baseline: CUTLASS SM100 FP16 GEMMs, tile-swept winners, split-G7 (sq tile) with GELU epilogue.
- Lands the MK-1 GeGLU megakernel + MK-2 F-2 bundle (mega geglu + G8 down-proj fused into one C entry), switchable via `G8_KERNEL` env (default `bundled`).
- Vendors the production sm100 GEMM kernel under `csrc/gemm/mega/cutlass/` (upstream `third_party/cutlass` untouched).
- 7 commits cherry-picked cleanly onto #35; the FP16 baseline (`use_fp8=False`) was already upstream.

## Results (Jetson AGX Thor, openpi-jax container)
- `g8_f2_cosine`: cos = 1.0000000000, max|diff| = 0 (bit-exact vs python chain)
- `stage_b_test`: hidden cos = 0.9999998808, PASS
- FP16 3-view: p50 85.4 ms; F-2 bundle = -0.94 ms vs unbundled two-launch (back-to-back A/B)

## Test plan
- [x] g8_f2_cosine production gate (cos = 1.0)
- [x] stage_b_test (PASS)
- [x] FP16 num_views=3 precision bench, n_warmup=15 n_trials=20 (no regression vs #35 baseline)
- [x] flash_rt_kernels builds clean on merged CMakeLists (motus + megakernel targets coexist)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,13 @@ else()
   set(ENABLE_SM100_CUTLASS OFF)
 endif()
 
+# Developer-only SM100 kernels. OFF by default so production flash_rt_kernels
+# neither compiles nor exports them; turn on locally for kernel bring-up.
+option(FLASHRT_BUILD_SM100_SWEEP
+       "Build the transient FP16 CUTLASS tile-sweep bench kernel" OFF)
+option(FLASHRT_BUILD_SM100_ENCODER_MLP
+       "Build the WIP Path C encoder MLP megakernel scaffold" OFF)
+
 # SM120+ (Blackwell): NVFP4/W4A8 block-scaled GEMM
 if(GPU_ARCH STREQUAL "120" OR GPU_ARCH STREQUAL "120a")
   set(ENABLE_NVFP4 ON)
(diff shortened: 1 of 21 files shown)


LiangSu8899:
Addressed the review in 8389e6d:

- **Dev-only kernels gated.** The transient FP16 tile-sweep bench (`cutlass_sm100_fp16_sweep`) and the WIP Path C encoder MLP scaffold (`encoder_mlp_sm100`) now build only behind `FLASHRT_BUILD_SM100_SWEEP` / `FLASHRT_BUILD_SM100_ENCODER_MLP` (both OFF by default). Their pybind entries are wrapped in the matching compile-time defines, so production `flash_rt_kernels` neither compiles, links, nor exports them — no unconditional `m.def` references a gated symbol.
- **Scaffold comments fixed.** `encoder_mlp_sm100.cu` now states its real (unbuilt-by-default, WIP) status; removed the dangling `megakernel_dev/notes/PATH_C_DESIGN.md` reference and other dead internal-doc/codename references.
- **Trailing whitespace removed** from the vendored CUTLASS and megakernel headers — `git diff --check` is clean against main.
- **Encoder env reads hoisted.** `_encoder_forward_fp16` reads `QKV_KERNEL` / `ENC_PATH` / `G8_KERNEL` / `O_KERNEL` / `USE_*` once per call (single module-level `import os`) instead of re-importing and re-reading inside the layer loop.

Re-validated on Jetson AGX Thor (SM110), default flags:
- `flash_rt_kernels` builds clean (exit 0); import check: production megakernel symbols present, `cutlass_fp16_sweep` / `encoder_mlp_fused_fp16` absent from the module.
- `g8_f2_cosine`: cos = 1.0000000000
- `stage_b_test`: hidden cos = 0.9999998808, PASS
- FP16 3-view (n_warmup 15 / n_trials 20): p50 85.7 ms (no regression vs prior)

SM89 is structurally unaffected: all gated targets sit inside `if(ENABLE_SM100_CUTLASS)` and the new bindings are inside `#ifdef ENABLE_SM100_CUTLASS`, so the SM89 build/import path is unchanged.

PR #40, opened 2026-05-21
fix(pi05): fix bf16 bias gelu binding shape

strayberry:
# Fix Pi0.5 BF16 Bias GELU Binding Shape

## Summary

This PR fixes the pybind wrappers for `bias_gelu_bf16` and
`bias_gelu_bf16_strict`.

Both wrappers already receive the logical BF16 activation matrix shape as
`(seq_len, dim)`, but they previously passed `(seq_len * dim, dim)` into
`bias_gelu_inplace_bf16`. That made the CUDA kernel operate on
`seq_len * dim * dim` elements instead of `seq_len * dim`, causing an
out-of-bounds write. In the Pi0.5 INT8 path, the failure surfaced later as a
cuBLAS error during vision encoder calibration.

## Change

Only the pybind argument forwarding is changed:

```diff
- seq_len * dim, dim
+ seq_len, dim
```

Affected bindings:

- `bias_gelu_bf16`
- `bias_gelu_bf16_strict`

No CUDA source, CMake target ownership, public API, or runtime routing is
changed.

## Root Cause

The CUDA helper expects a matrix shape:

```cpp
bias_gelu_inplace_bf16(x, bias, M, N, stream)
```

The Python binding receives:

```cpp
seq_len, dim
```

So the correct call is:

```cpp
bias_gelu_inplace_bf16(..., seq_len, dim, ...)
```

For Pi0.5 vision FFN, a representative INT8 activation shape is:

```text
seq_len = 512
dim = 4304
```

The old wrapper therefore described the buffer as:

```text
M = 512 * 4304
N = 4304
```

That is much larger than the allocated activation tensor.

## Why This Was Observed In INT8

The default BF16 Pi0.5 path does not use this fused wrapper. It uses separate
bias add and GELU calls:

```python
_bias_add_bf16(...)
gelu_inplace(...)
```

The full Pi0.5 INT8 path uses the fused BF16 activation wrapper after the
vision FFN up GEMM:

```python
fvk.bias_gelu_bf16_strict(..., seq, VIS_H, ...)
```

That is why the issue was exposed by INT8 testing even though the faulty
kernel operates on BF16 activation buffers.

## Validation Environment

**Hardware & System**

| Field | Value |
|---|---|
| Device | Jetson AGX Orin 32G |
| Machine | KST Jetson AGX Orin AVSAI |
| SoC | tegra234 |
| Arch | aarch64 |
| RAM | 30698 MB |
| Visible GPU memory | 29.98 GB |
| SM count | 14 |
[...]

PR #41, opened 2026-05-21
Revert "fix(pi05): fix bf16 bias gelu binding shape"

LiangSu8899:
Reverts LiangSu8899/FlashRT#40

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,6 +118,29 @@ Every pybind entry must have matching CMake target ownership.
   and the target that already had them does not fail with duplicate
   definitions.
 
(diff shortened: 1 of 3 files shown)


LiangSu8899:
Added integration checklist docs to prevent regressions like #30.

The new guidance makes kernel alias/fused-helper refactors explicitly check:
- legacy pybind ABI and argument mapping
- shape contracts such as (seq_len, dim) vs (M, N)
- dtype / strict rounding / in-place semantics
- optional fast paths vs required hardware routing

#40 is referenced as the concrete fix pattern for the #30 bias_gelu_bf16(_strict) alias regression.

PR #48, opened 2026-05-23
feat(pi05): pixel-MAE adaptive refresh for temporal KV cache

strayberry:
## Summary

Add adaptive refresh to Pi0.5 RTX temporal KV cache — the cached encoder K/V can now be flushed early based on an external signal (e.g., pixel-level scene change detection), not just on a fixed schedule. This preserves the latency benefit of cache=N while improving action quality when the visual input changes between scheduled full frames.

## Changes

- **`flash_rt/frontends/torch/pi05_rtx.py`** — `infer()` gains `force_full` parameter and returns `used_full_pipeline` / `cache_forced_full` flags. Frame-count reset logic ensures a forced refresh starts a new reuse window. No change to the fixed-window scheduling math when `force_full` is not used.
- **`examples/orin/eval_libero.py`** — new offline benchmark script that auto-downloads a pre-exported LIBERO NPZ (42 MB, 3 episodes × 100 frames, exported via [export_libero_npz.py](https://github.com/strayberry/mm-edge-infer-accel/blob/main/scripts/export_libero_npz.py)) and runs 4 configs (baseline, cache2, cache3, adaptive_cache3) reporting latency and cosine similarity vs baseline. The NPZ is a pre-exported snapshot of LeRobotDataset frames, avoiding the LeRobot/SIMEP runtime dependency and enabling reproducible offline eval without reloading the dataset.

## Prior experiment: full sweep

All 7 configurations tested offline on 300 pre-exported LIBERO observation frames. BF16, threshold=6.5 for adaptive configs.

| Config | p50 | forced_full | cos_mean | cos_min | Retained? |
|---|---|---|---|---|---|
| bf16_baseline (cache=1) | 218.3 ms | — | baseline | baseline | yes |
| bf16_cache2 | 135.5 ms | — | 0.982403 | 0.478313 | **yes** — highest fidelity, 7.4 Hz |
| bf16_cache3 | 56.6 ms | — | 0.960540 | 0.122606 | yes |
| bf16_cache4 | 55.9 ms | — | 0.945754 | −0.058609 | no — diminishing returns, tail quality drops |
| bf16_adaptive_cache2 | 216.4 ms ❌ | 34/300 | 0.983152 | 0.478313 | no — p50 jumps to full-frame latency |
| **bf16_adaptive_cache3** | **55.6 ms** ✅ | **89/300** | **0.974735** | **0.434654** | **yes** — best speed/quality trade-off |
| bf16_adaptive_cache4 | 56.6 ms | 106/300 | 0.964295 | −0.058609 | no — mean improves but tail doesn't |

Key finding: `adaptive_cache3` is the only clear win — maintains cache3 p50 (~56 ms) while significantly improving quality (cos_mean 0.961→0.975, cos_min 0.123→0.435). `adaptive_cache2` is unusable (p50 jumps to full-frame); `adaptive_cache4` improves mean but not tail. Fixed cache4 and all adaptive_cache2/4 are not retained as main tracks.

## Benchmark (Jetson AGX Orin, 300 frames, BF16)

Final numbers from the PR-branch build:

| Config | p50 | p95 | forced_full | cos_mean | cos_min |
|---|---|---|---|---|---|
| bf16_baseline (cache=1) | 218.4 ms | 218.7 ms | — | baseline | baseline |
| bf16_cache2 | 135.7 ms | 216.8 ms | — | 0.981 | 0.556 |
| bf16_cache3 | 55.7 ms | 217.2 ms | — | 0.959 | −0.371 |
| **bf16_adaptive_cache3** | **55.4 ms** | **216.6 ms** | **91/300** | **0.974** | **0.534** |

Results consistent with prior experiment. The adaptive refresh triggers on ~30% of frames (91/300) at the default pixel-MAE threshold of 6.5, catching scene transitions that fixed-window cache3 misses.

### Threshold sensitivity (adaptive_cache3, 30 frames)

| Threshold | p50 | forced_full | cos_mean | cos_min |
|---|---|---|---|---|
| 5.0 | 216.8 ms ❌ | 16/30 | 0.996 | 0.896 |
| 5.5 | 215.1 ms ❌ | 11/30 | 0.995 | 0.896 |
| 6.0 | 57.2 ms ✅ | 9/30 | 0.995 | 0.889 |
| **6.5** | **55.6 ms** ✅ | **7/30** | **0.998** | **0.992** |
| **7.0** | **55.6 ms** ✅ | **7/30** | **0.998** | **0.992** |
| 7.5 | 55.4 ms | 5/30 | 0.994 | 0.896 ❌ |

The default threshold of 6.5 sits in a stable operating window (6.0–7.0). Below 6.0 the cache refreshes too aggressively, collapsing p50 to full-frame latency. At 7.5 the sweep begins to miss scene transitions (cos_min drops to 0.896), making it unreliable. Thresholds 6.5 and 7.0 produced identical results on this 30-frame sample, confirming the metric is not critically sensitive to the exact value within the operating window.

> **Note:** The optimal threshold is data-dependent and may vary with camera placement, scene complexity, and task characteristics. The pixel-MAE distribution changes with resolution, lighting, and background complexity — deployment on different environments or robot setups may require adjusting this value.

### Retained cache routes

| Route | Use case |
|---|---|
| **bf16_cache2** | highest fidelity (cos_mean=0.981), p50=136ms, 7.4 Hz |
| **bf16_adaptive_cache3** | best speed/quality trade-off (cos_mean=0.974), p50=55ms, ~18 Hz |

## Reproduce

```bash
# Quick smoke test (30 frames, ~2 min)
python examples/orin/eval_libero.py \
    --checkpoint /path/to/pi05_libero_finetuned_v044

# Full benchmark (300 frames, ~12 min)
python examples/orin/eval_libero.py \
    --checkpoint /path/to/pi05_libero_finetuned_v044 \
    --frames 300

# Threshold sensitivity
for t in 6.0 6.5 7.0; do
    python examples/orin/eval_libero.py \
        --checkpoint /path/to/ckpt --frames 30 \
        --threshold $t --configs bf16_adaptive_cache3
done

The NPZ evaluation dataset downloads automatically on first run (~42 MB).
```

## Test plan

- [x] Quick test: `python examples/orin/eval_libero.py --checkpoint /path/to/ckpt` (30 frames, ~2 min)
- [x] Full eval: `python examples/orin/eval_libero.py --checkpoint /path/to/ckpt --frames 300` (~12 min)
- [x] Verified fixed-window cache scheduling is bit-identical to original when `force_full=False`
- [x] `_infer_cfg_batched()` path unaffected (no temporal cache logic)

### DIFF
--- a/examples/orin/eval_libero.py
+++ b/examples/orin/eval_libero.py
@@ -1,0 +1,235 @@
+#!/usr/bin/env python3
+"""
+FlashRT Orin — LIBERO action similarity benchmark.
+
+Offline comparison of action chunks across cache configurations using
+pre-exported LIBERO observation frames. Downloads the evaluation dataset
+automatically on first run (42 MB).
+
+Default --frames=30 completes in ~30 seconds per config for a quick
+smoke test. Use --frames=300 for the full evaluation (~3 min per config,
+stable latency/cosine metrics for PR validation).
+
+Usage:
+    python examples/orin/eval_libero.py \
+        --checkpoint /path/to/pi05_libero_finetuned_v044
+
+    python examples/orin/eval_libero.py \
+        --checkpoint /path/to/pi05_libero_finetuned_v044 \
+        --configs bf16_baseline,bf16_cache2,bf16_adaptive_cache3
+"""
+
+import argparse
+import gc
+import logging
+import os
+import statistics
+import sys
+import time
+import urllib.request
+from pathlib import Path
+
+import numpy as np
+import torch
+from PIL import Image
+
+REPO = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(REPO))
+
+from flash_rt.frontends.torch.pi05_rtx import Pi05TorchFrontendRtx  # noqa: E402
+
+logging.basicConfig(level=logging.WARNING)
+
+NPZ_URL = ("https://github.com/strayberry/mm-edge-infer-accel/releases/"
+           "download/v0.1.0/libero_episodes0_1_2_100frames_each.npz")
+CACHE_DIR = Path.home() / ".flash_rt" / "bench_data"
+BASELINE = "bf16_baseline"
+
+CONFIGS = {
+    BASELINE:               dict(cache=1),
+    "bf16_cache2":          dict(cache=2),
+    "bf16_cache3":          dict(cache=3),
+    "bf16_adaptive_cache3": dict(cache=3, refresh="pixel_delta"),
+}
+
+
+def parse_args():
+    p = argparse.ArgumentParser(description="FlashRT Orin LIBERO action similarity benchmark")
+    p.add_argument("--checkpoint", "-c", required=True)
+    p.add_argument("--npz", default="",
+                   help="Path to NPZ (auto-downloads if empty)")
+    p.add_argument("--frames", type=int, default=30,
+                   help="Frames to evaluate (0 = all, 300 for full eval)")
+    p.add_argument("--warmup", type=int, default=3)
+    p.add_argument("--seed", type=int, default=1234)
+    p.add_argument("--threshold", type=float, default=6.5,
+                   help="Pixel-MAE refresh threshold for adaptive cache")
+    p.add_argument("--configs",
+                   default="bf16_baseline,bf16_cache2,bf16_cache3,bf16_adaptive_cache3",
+                   help="Comma-separated config names")
+    return p.parse_args()
+
+
+def resize_images(images, size=224):
+    if images.shape[1:3] == (size, size):
+        return images
+    resized = []
+    for img in images:
+        resized.append(np.asarray(Image.fromarray(img).resize((size, size), Image.BILINEAR)))
+    return np.stack(resized).astype(np.uint8)
+
+
+def load_npz(npz_path, max_frames):
+    data = np.load(npz_path, allow_pickle=True)
+    n = max_frames if max_frames > 0 else len(data["images"])
+    return {
+        "images": resize_images(np.asarray(data["images"][:n])),
+        "wrist_images": resize_images(np.asarray(data["wrist_images"][:n])),
+    }
+
+
+def download_npz(target):
+    print(f"Downloading evaluation dataset...", flush=True)
+    target.parent.mkdir(parents=True, exist_ok=True)
+    urllib.request.urlretrieve(NPZ_URL, target)
+    print(f"Saved to {target}", flush=True)
+    return str(target)
+
+
+def summarize(values):
+    vals = sorted(float(x) for x in values)
+    if not vals:
+        return {}
+    return {
+        "mean": float(np.mean(vals)),
+        "p50": statistics.median(vals),
+        "p95": vals[int(len(vals) * 0.95)],
+        "min": vals[0],
+        "max": vals[-1],
+    }
+
+
+def cosine(a, b):
+    x = a.reshape(-1).astype(np.float32)
+    y = b.reshape(-1).astype(np.float32)
+    denom = np.linalg.norm(x) * np.linalg.norm(y)
+    return 0.0 if denom == 0 else float(np.dot(x, y) / denom)
+
+
+def pixel_delta_mae(left, right):
+    return float(np.mean(np.abs(left.astype(np.int16) - right.astype(np.int16))))
+
+
+def main():
+    args = parse_args()
+    requested = [n.strip() for n in args.configs.split(",") if n.strip()]
+    unknown = [n for n in requested if n not in CONFIGS]
+    if unknown:
+        raise SystemExit(f"Unknown configs: {unknown}")
+    if BASELINE not in requested:
+        requested.insert(0, BASELINE)
+
+    # Resolve NPZ
+    npz_path = args.npz or str(CACHE_DIR / Path(NPZ_URL).name)
+    if not os.path.isfile(npz_path):
+        npz_path = download_npz(Path(npz_path))
+
+    data = load_npz(npz_path, args.frames)
+    images = data["images"]
+    wrist_images = data["wrist_images"]
+    n = len(images)
+
+    print("=" * 55)
+    print("FlashRT Orin — LIBERO action similarity benchmark")
+    print(f"  Frames:  {n}")
+    print(f"  Seed:    {args.seed}")
+    print(f"  Configs: {', '.join(requested)}")
+    print("=" * 55)
+
+    all_actions = {}
+    all_lat = {}
+    prompt = "pick up the object"
+
+    for name in requested:
+        cfg = CONFIGS[name]
+        use_adaptive = cfg.get("refresh") == "pixel_delta"
+        print(f"\n  --- {name}: cache={cfg['cache']}" +
+              (" adaptive" if use_adaptive else ""))
+
+        pipe = Pi05TorchFrontendRtx(
+            args.checkpoint, num_views=2, num_steps=10,
+            vision_pool_factor=1, vision_num_layers=27,
+            cache_frames=cfg["cache"],
+        )
+
+        torch.manual_seed(args.seed)
+        torch.cuda.manual_seed_all(args.seed)
+        pipe.set_prompt(prompt)
+        pipe.calibrate_with_real_data([{
+            "image": images[0], "wrist_image": wrist_images[0],
+        }])
+
+        for i in range(min(args.warmup, n)):
+            torch.manual_seed(args.seed + i)
+            torch.cuda.manual_seed_all(args.seed + i)
+            pipe.infer({"image": images[i], "wrist_image": wrist_images[i]})
+
+        pipe.set_prompt(prompt)
+        actions = [None] * n
+        lat_ms = np.empty((n,), dtype=np.float32)
+        forced = np.zeros((n,), dtype=np.bool_)
+        last_full_image = last_full_wrist = None
+
+        for i in range(n):
+            torch.manual_seed(args.seed + i)
+            torch.cuda.manual_seed_all(args.seed + i)
+            obs = {"image": images[i], "wrist_image": wrist_images[i]}
+
+            force = False
+            if use_adaptive and last_full_image is not None:
+                delta = max(
+                    pixel_delta_mae(images[i], last_full_image),
+                    pixel_delta_mae(wrist_images[i], last_full_wrist),
+                )
+                force = delta >= args.threshold
+
+            t0 = time.perf_counter()
+            out = pipe.infer(obs, force_full=force)
+            lat_ms[i] = (time.perf_counter() - t0) * 1000.0
+            actions[i] = np.asarray(out["actions"], dtype=np.float32)
+
+            if out.get("used_full_pipeline", True):
+                last_full_image = images[i]
+                last_full_wrist = wrist_images[i]
+            forced[i] = out.get("cache_forced_full", False)
+
+        all_actions[name] = np.stack(actions)
+        all_lat[name] = lat_ms
+        s = summarize(lat_ms)
+        forced_count = int(np.sum(forced))
+        extra = f"  forced_full={forced_count}/{n}" if use_adaptive else ""
+        print(f"    p50={s['p50']:.1f}ms  p95={s['p95']:.1f}ms  "
+              f"mean={s['mean']:.1f}ms{extra}")
+
+        del pipe
+        gc.collect()
+        torch.cuda.empty_cache()
+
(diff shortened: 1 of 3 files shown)


LiangSu8899:
Thanks for the contribution, this is a very interesting direction.

One thing I want to be careful about: in my previous Pi0.5 experiments, almost every form of temporal / KV cache reuse caused task-level degradation, even when some offline metrics looked acceptable. I also observed a pretty unstable cosine cliff phenomenon: depending on the input frame, the reference action similarity could suddenly diverge a lot.

As a comparison, OpenVLA-style architectures seem much more tolerant to cache manipulation. My current guess is that this may come from architectural differences in the attention path: Pi0.5 has a tighter vision-language/action coupling through the diffusion/action decoder, so stale visual KV states can directly perturb the action distribution, while OpenVLA’s autoregressive token path may absorb or localize this kind of cache error better. But I’m not fully sure yet.

So I’m very curious here: do you have any closed-loop sim benchmark or real-robot comparison for this adaptive refresh strategy? For example, LIBERO success rate with cache=1 vs cache=3 vs adaptive cache3 would be much more convincing than offline action cosine alone.

--To be clear, a small accuracy drop is not necessarily unacceptable for this feature. What I’m mainly trying to understand is the distribution across a broader test range: does the degradation stay smooth and bounded, or are there unstable cliff cases? In my own tests, this kind of cache reuse showed very input-dependent instability.

Thx!

strayberry:
> Thanks for the contribution, this is a very interesting direction.
> 
> One thing I want to be careful about: in my previous Pi0.5 experiments, almost every form of temporal / KV cache reuse caused task-level degradation, even when some offline metrics looked acceptable. I also observed a pretty unstable cosine cliff phenomenon: depending on the input frame, the reference action similarity could suddenly diverge a lot.
> 
> As a comparison, OpenVLA-style architectures seem much more tolerant to cache manipulation. My current guess is that this may come from architectural differences in the attention path: Pi0.5 has a tighter vision-language/action coupling through the diffusion/action decoder, so stale visual KV states can directly perturb the action distribution, while OpenVLA’s autoregressive token path may absorb or localize this kind of cache error better. But I’m not fully sure yet.
> 
> So I’m very curious here: do you have any closed-loop sim benchmark or real-robot comparison for this adaptive refresh strategy? For example, LIBERO success rate with cache=1 vs cache=3 vs adaptive cache3 would be much more convincing than offline action cosine alone.
> 
> --To be clear, a small accuracy drop is not necessarily unacceptable for this feature. What I’m mainly trying to understand is the distribution across a broader test range: does the degradation stay smooth and bounded, or are there unstable cliff cases? In my own tests, this kind of cache reuse showed very input-dependent instability.
> 
> Thx!

Thanks for the detailed feedback. I agree that offline cosine alone is not enough for Pi0.5 cache reuse, especially given the unstable cliff behavior you pointed out.

I added and ran `examples/orin/eval_libero.py`, a Thor-style Orin closed-loop LIBERO check that follows the protocol in `examples/thor/eval_libero.py` as closely as possible, with the only intended changes being the Orin/Pi0.5 frontend and cache settings.

On a full `libero_spatial` sweep with all 10 tasks x 3 trials:

| Config | Success | Used full pipeline | Forced full | p50 latency |
| --- | ---: | ---: | ---: | ---: |
| BF16 cache=1 | 24/30 | 807/807 | - | ~219 ms |
| BF16 cache=2 | 22/30 | 446/884 | - | ~186 ms |
| BF16 cache=3 | 11/30 | 402/1176 | - | ~57 ms |
| BF16 adaptive cache=3 | 21/30 | 864/867 | 834/867 | ~220 ms |

Here, `used full pipeline` counts replans that actually executed the full vision/encoder path instead of the cache-only path. `forced full` only counts early refreshes requested by the adaptive policy; scheduled full frames are counted in `used full pipeline` but not in `forced full`.

This confirms your concern on a broader sweep. Fixed cache=3 is very fast but causes substantial task-level degradation. The current pixel-MAE adaptive refresh recovers much of that success rate, but it is too conservative in closed-loop: it forced a full refresh on 834/867 replans and used the full pipeline on 864/867 replans, so the latency benefit mostly disappears.

BF16 cache=2 remains the more conservative cache route: it is close to the cache=1 success rate on this run, but the speedup is much smaller than cache=3 because roughly half of the replans still run the full pipeline.

Given these results, I do not think the current pixel-MAE adaptive refresh should be presented as a default recommendation. I’m happy to close this PR if you think that is the cleanest path, or continue the discussion if you think some smaller part of the mechanism is still useful.

I also tried a few other frame-level variants. Lower pixel-MAE thresholds over-refresh and move p50 back toward full-frame latency, while higher thresholds miss cliff cases. A broad stage-based refresh improves similarity, but also moves p50 close to the BF16 baseline. cache=4/adaptive cache=4 improves the mean offline cosine compared with fixed cache=4, but the tail remains poor (`cos_min < 0`). Together with the closed-loop result above, this suggests that frame-level refresh is not selective enough for this Pi0.5 path: it either under-refreshes and keeps cache cliffs, or over-refreshes and loses the cache speedup.

That said, I still think vision-token reuse is a promising direction for accelerating VLA inference. This also made me look into the VLA-Cache work on adaptive token caching:

VLA-Cache: Efficient Vision-Language-Action Manipulation via Adaptive Token Caching, Siyu Xu, Yunke Wang, Chenghao Xia, Dihao Zhu, Tao Huang, Chang Xu, https://arxiv.org/abs/2502.02175

Of course, this work is mainly on OpenVLA-style architectures, and your point about the possible architectural differences between Pi0.5 and OpenVLA still applies, so I think this direction needs to be re-validated on the Pi0.5 pipeline.

As a first check, I wanted to see whether there is enough temporal redundancy at the visual-token level to make partial refresh worth exploring. I used the 300-frame offline LIBERO NPZ and split each 224x224 camera view into a 16x16 spatial grid. With two camera views, this gives 512 patch cells per frame. For each cache reuse frame, I compared it with the previous full-refresh frame and counted how many patch cells changed above a given pixel-MAE threshold.

| Cache window | Patch threshold | Median changed patch cells | Median changed ratio | Frames with <=64 changed patch cells |
| --- | ---: | ---: | ---: | ---: |
| cache=2 | 15 | 32 / 512 | 6.2% | 99.3% |
| cache=2 | 20 | 22 / 512 | 4.3% | 100% |
| cache=3 | 20 | 29 / 512 | 5.7% | 96.5% |
| cache=4 | 20 | 35 / 512 | 6.8% | 91.1% |

This means most reuse frames only have a small number of visibly changed patch cells. For example, with cache=2 and threshold=20, the median frame changes only 22 out of 512 patch cells, and all reuse frames stay within 64 changed patch cells. So from a compute-budget perspective, partial token refresh looks plausible: in many frames, only a small subset of visual regions changes.

However, this does not mean raw RGB patch change is a good refresh signal. I also checked whether these image-space changes correlate with action error. I aligned the 300-frame NPZ with the cache=2 per-frame cosine CSV, kept only cache=2 reuse frames, and compared image-space metrics against `cosine_loss = 1 - cosine_vs_full_baseline`.

The correlations were almost zero:

| Metric | Spearman vs cosine loss |
| --- | ---: |
| whole-image pixel MAE | 0.030 |
| top-4 patch MAE | 0.045 |
| top-16 patch MAE | 0.004 |

One representative cache cliff frame had:

```text
cosine_vs_full_baseline = 0.478
raw whole-image pixel MAE = 1.27
changed patch cells @ threshold 15 = 0 / 512
changed patch cells @ threshold 20 = 0 / 512
```

So this probe suggests two things: partial token refresh has compute-budget potential, but raw RGB image-space MAE is not a reliable action-cliff predictor for Pi0.5. A useful token-level policy likely needs a more model-aware signal, such as feature-space token delta or task/action-aware token importance.

My current interpretation is:

1. Frame-level pixel-MAE refresh is not a strong enough closed-loop policy for Pi0.5.
2. The data supports your observation that Pi0.5 cache reuse has input-dependent cliff cases.
3. I do not want to scope a full partial-KV/token-reuse pipeline into this PR yet, because the current Pi0.5 RTX pipeline only has full and decoder-only paths.

I plan to explore this token-level direction separately, and I’d be happy to discuss it further if you are interested.

LiangSu8899:
Thanks a lot for the detailed experiments. I think this PR is valuable as a research direction, but at the current stage I’d prefer not to merge it as a visible/default Pi0.5 feature yet.

Your closed-loop results basically confirm the main concern: fixed cache reuse can give large latency gains, but the task-level degradation is not smooth enough; adaptive refresh recovers part of the success rate, but becomes too conservative and loses most of the speedup.

This also matches my previous Pi0.5 experiments. I tried many similar reuse strategies before: temporal KV reuse, input-token reuse, visual-region change thresholds, and several frame/object-level heuristics. Almost all of them eventually showed cliff-like accuracy drops on some inputs. Interestingly, simply reducing diffusion/action steps often had better tolerance than cache reuse, although on real robots that can also introduce uneven instability.

So my current feeling is: for Pi0.5, this is probably not ready to be exposed as a user-facing feature yet, especially on edge devices where people may expect it to be reliable. But the direction is definitely important.

One path that may be more practical is step-reduction distillation for Pi0.5, instead of directly reducing steps or reusing stale cache. That might preserve behavior better while still reducing latency.

There are also a few reuse directions I haven’t fully tested yet:

* feature-space token delta instead of raw RGB pixel delta
* action-aware token importance / refresh policy
* confidence-based early refresh using decoder hidden states

So I’d prefer to archive/close this PR rather than merge it, but I’d be very happy to keep discussing this direction. I think your experiments are useful, and this is exactly the kind of edge-side VLA optimization problem worth exploring further.

strayberry:
> Thanks a lot for the detailed experiments. I think this PR is valuable as a research direction, but at the current stage I’d prefer not to merge it as a visible/default Pi0.5 feature yet.
> 
> Your closed-loop results basically confirm the main concern: fixed cache reuse can give large latency gains, but the task-level degradation is not smooth enough; adaptive refresh recovers part of the success rate, but becomes too conservative and loses most of the speedup.
> 
> This also matches my previous Pi0.5 experiments. I tried many similar reuse strategies before: temporal KV reuse, input-token reuse, visual-region change thresholds, and several frame/object-level heuristics. Almost all of them eventually showed cliff-like accuracy drops on some inputs. Interestingly, simply reducing diffusion/action steps often had better tolerance than cache reuse, although on real robots that can also introduce uneven instability.
> 
> So my current feeling is: for Pi0.5, this is probably not ready to be exposed as a user-facing feature yet, especially on edge devices where people may expect it to be reliable. But the direction is definitely important.
> 
> One path that may be more practical is step-reduction distillation for Pi0.5, instead of directly reducing steps or reusing stale cache. That might preserve behavior better while still reducing latency.
> 
> There are also a few reuse directions I haven’t fully tested yet:
> 
> * feature-space token delta instead of raw RGB pixel delta
> * action-aware token importance / refresh policy
> * confidence-based early refresh using decoder hidden states
> 
> So I’d prefer to archive/close this PR rather than merge it, but I’d be very happy to keep discussing this direction. I think your experiments are useful, and this is exactly the kind of edge-side VLA optimization problem worth exploring further.


Thanks a lot for the thoughtful feedback. The step-reduction distillation direction is very interesting, and I agree that it may be a more promising path for Pi0.5 latency reduction than directly reducing steps. It sounds like a cleaner way to reduce denoising cost while keeping the policy behavior closer to the original model.

I also did a few quick offline checks for the three follow-up reuse directions you mentioned:

| Direction | Initial result |
| --- | --- |
| Feature-space token delta | `vision_x` deltas work better than RGB patch MAE (`vision_x_top16` AUROC `0.735` for `cos < 0.8`, vs RGB top-16 patch MAE `0.384`). The catch is that `vision_x` already requires the vision encoder, so using it directly at runtime would give back much of the cache speedup. |
| Action-aware token importance | The signal is weak, but action-specific errors are not completely uniform. Translation error correlates more with wrist-view token deltas than main-view token deltas; rotation/gripper are less clear. This is only a correlation clue, not a refresh policy yet. |
| Decoder/action confidence | I have not tested decoder hidden states directly yet, but a simple output-level signal looks more promising: `action_step_delta_mean` has Spearman `0.603` with cache2 cosine loss and AUROC `0.896` for `cos < 0.8`. |

The output-confidence result suggests a possible speculative fallback direction:

```text
run cache-only decoder -> check action/chunk confidence -> rerun full pipeline only if abnormal
```

This would still need a threshold/fallback-rate study because fallback frames cost `cache-only + full` latency, but it looks more practical than frame-level refresh since it does not require running the vision encoder just to decide whether to refresh.

I will close this PR for now and continue exploring other directions in separate branches. Thanks again for the detailed discussion.

PR #62, opened 2026-05-30
feat(async_loader): add camera_map parameter to make_step_dataloader

eliasab16:
While finetuning the model, I had a camera naming issues because the default libero cameras were hardcoded, so it wasn't possible to finetune on my dataset without patching the code.

This pr adds camera_map parameter to make_step_dataloader. If not provided, the code falls back to the libero default.

### DIFF
--- a/tests/test_recap_async_camera_map.py
+++ b/tests/test_recap_async_camera_map.py
@@ -1,0 +1,87 @@
+import sys
+import types
+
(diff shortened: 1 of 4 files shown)


LiangSu8899 (approved):

LiangSu8899:
Thanks a lot for this fix!

PR #86, opened 2026-06-11
Pi0.5 RTX: fixed-shape state-prompt graph option (one graph, no recap…

LiangSu8899:
…ture)

Pi0.5 renders discretized robot state into the prompt, so the token length drifts with state values. Add a `state_prompt_mode` to the RTX Pi0.5 frontend:

  - "exact" (default, unchanged behavior): a separate pipeline captured per exact prompt length, cached; pair with warm_state_prompt_buckets() to front-load the lengths you expect.
  - "fixed" (opt-in): ONE pipeline + ONE captured graph at the max prompt length serves every length. Padded prefix keys are masked via FlashAttention-2 seqused_k; the decoder appends its action K/V right after the valid prefix through a new qkv_split_rope_devpos kernel, so a changing state length never re-captures a graph or reruns autotune and no warmup is needed. Select via state_prompt_mode="fixed" or FLASHRT_PI05_STATE_PROMPT_MODE=fixed.

Validated on Pi0.5 (RTX 5090): BF16 fixed-vs-exact cosine >= 0.9999 over a varying-length state sequence (seqused + devpos reproduce the per-length computation bit-for-bit), one captured graph vs per-length, zero recapture. Non-state prompts are unchanged.

New additive kernel qkv_split_rope_devpos (rope.cu + binding); existing kernels untouched.

### 2. Per-frame precision on real LIBERO frames (8 frames, real state, identical noise)

Unnormalized robot actions; `original` = unquantized BF16 model output.

| frame | len | fixed-FP8 vs original | exact-FP8 vs original | fixed-FP8 vs exact-FP8 |
|---|---|---|---|---|
|  |  | cos / max_abs | cos / max_abs | cos / max_abs |
| 0 | 62 | 0.999981 / 0.0078 | 0.999987 / 0.0065 | 0.999978 / 0.0081 |
| 1 | 62 | 0.999976 / 0.0117 | 0.999988 / 0.0078 | 0.999978 / 0.0099 |
| 2 | 62 | 0.999802 / 0.0395 | 0.999764 / 0.0428 | 0.999982 / 0.0082 |
| 3 | 60 | 0.999901 / 0.0154 | 0.999967 / 0.0147 | 0.999917 / 0.0152 |
| 4 | 61 | 0.999964 / 0.0194 | 0.999987 / 0.0117 | 0.999960 / 0.0162 |
| 5 | 61 | 0.999962 / 0.0147 | 0.999975 / 0.0156 | 0.999963 / 0.0156 |
| 6 | 61 | 0.999891 / 0.0280 | 0.999890 / 0.0321 | 0.999953 / 0.0195 |
| 7 | 61 | 0.999976 / 0.0109 | 0.999969 / 0.0195 | 0.999956 / 0.0234 |
| **mean** |  | **0.999932** | **0.999941** | **0.999961** |

**fixed-FP8 vs original (0.99993) ≈ exact-FP8 vs original (0.99994)** — fixed mode
is as accurate as the shipped exact path against the unquantized model; the
difference is FP8 noise, not a regression.

## Latency (RTX 5090, FP8, pure CUDA-graph replay)

Measured as **pure graph replay** (CUDA events around the captured graph, 500-iter
warmup); the replay is deterministic (std ≤ 0.03 ms). Prompt ≈ 51 tokens,
`PI05_STATE_PROMPT_MAX_LEN=200`.

| views | `exact` | `fixed` | **fixed − exact** |
|---|---|---|---|
| 2 | 18.68 ms | 23.21 ms | **+4.53 ms** |
| 3 | 21.37 ms | 28.10 ms | **+6.74 ms** |
| **3 − 2 (extra view)** | +2.69 ms | +4.90 ms | |

- `"fixed"` runs **every** inference at the padded max length (200 here vs the
  ~51-token prompt), so it is ~+4.5 ms (2v) / +6.7 ms (3v) slower than `"exact"`
  at that length — the price for never re-capturing as the length drifts. The
  gap shrinks as `PI05_STATE_PROMPT_MAX_LEN` approaches the real maximum.
- The extra camera view costs +2.69 ms (`exact`) / +4.90 ms (`fixed`); `fixed`
  pays more because the third view's tokens add onto the larger padded base.
- A full `predict()` adds ~1–1.6 ms over pure replay (noise gen + image H2D +
  output D2H) — that overhead is identical in both modes.
- Recommendation for peak latency: use `"exact"` and warm a fuller length sweep
  with `warm_state_prompt_buckets()`; use `"fixed"` when you'd rather not
  enumerate lengths and can spend the padding overhead.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -224,6 +224,7 @@ model = flash_rt.load_model(
 | `vision_pool_factor` | `int\|None` | `None` | Pi0.5 torch RTX/Orin. Spatial pooling factor for vision tokens. The FP16 RTX path currently supports only `1`. |
 | `vision_num_layers` | `int\|None` | `None` | Pi0.5 torch RTX/Orin. Number of SigLIP vision layers to run. |
 | `cache_frames` | `int\|None` | `None` | Pi0.5 torch RTX/Orin. Temporal encoder K/V cache period; `1` means no temporal reuse. |
+| `state_prompt_mode` | `str` | `"exact"` | Pi0.5 torch RTX. State-in-prompt graph strategy: `"exact"` (per-length capture + `warm_state_prompt_buckets()`) or `"fixed"` (one graph at the max length, no recapture; ~+4.6 ms/2v, +6.7 ms/3v). Env override `FLASHRT_PI05_STATE_PROMPT_MODE`. See [Pi0.5 State Prompts](#pi05-state-prompts). |
 
 ### Pi0.5 state prompt bucket warmup
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -139,8 +139,12 @@ ONE pipeline and ONE captured CUDA Graph at the max prompt length serve every
 length: the padded prefix keys are masked (FlashAttention-2 `seqused_k`) and the
 decoder's action K/V are appended right after the *valid* prefix
 (`qkv_split_rope_devpos`), so a changing state-token length **never re-captures
-a graph or reruns autotune** — no warmup needed. Enable it when you don't want
-to enumerate state lengths up front:
+a graph or reruns autotune** — no warmup needed. The trade-off is that every
+inference runs at the padded max length, so steady-state latency is somewhat
+higher than `"exact"` at a given length (it buys you no recapture stalls when
+the length drifts). Lower `PI05_STATE_PROMPT_MAX_LEN` toward your real maximum
+to shrink the padding. Enable it when you don't want to enumerate state lengths
+up front:
 
 ```python
 model = flash_rt.load_model(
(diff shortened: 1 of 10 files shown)


LiangSu8899:
The trade-off is latency:
every inference runs at the padded max length, so `"fixed"` is ~+4.6 ms (2
views) / +6.7 ms (3 views) slower than `"exact"` at a typical prompt length —
for peak performance prefer `"exact"` and warm a fuller length sweep yourself
with `warm_state_prompt_buckets()`. (Lowering `PI05_STATE_PROMPT_MAX_LEN`
toward your real maximum shrinks the padding and the gap.) Enable `"fixed"`
when you don't want to enumerate state lengths up front:

PR #104, opened 2026-06-22
docs: update JAX Orbax install notes

heiheiha798:
## Summary

- update the optional JAX frontend install notes after the Orbax StepMetadata loader support landed
- replace the stale statement that FlashRT expects only the Orbax 0.5.x metadata layout
- point future Orbax pin changes at the existing StepMetadata regression test

## Validation

Environment:
- CPU/static docs validation only
- Python: `/data/home/tianjianyang/miniconda3/envs/flashrt/bin/python`

Commands:

```bash
/data/home/tianjianyang/miniconda3/envs/flashrt/bin/python -m pytest tests/test_weight_loader.py -q
git diff --check -- docs/INSTALL.md
```

Results:

```text
20 passed in 1.91s
git diff --check passed
```

## Limitations / Not Covered

- Docs-only change; no runtime code, kernels, CMake, or pybind bindings changed.
- No GPU, latency, or precision validation was run because this only updates install guidance.
- No JAX / Orbax version pin is changed in this PR.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -183,14 +183,16 @@ below are what we test against — don't upgrade one without the others:
 pip install jax==0.5.3 jax-cuda12-pjrt==0.5.3 jax-cuda12-plugin==0.5.3 ml_dtypes==0.5.3 orbax-checkpoint flax
 ```
 
-Upgrade path (tracked, not yet done):
+Version drift notes:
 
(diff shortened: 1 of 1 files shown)


heiheiha798:
Closing this PR because the proposed docs-only update is too small to justify a standalone contribution. I will revisit if there is a more substantive related change.

PR #110, opened 2026-06-25
Add SM89 Qwen3-VL official FP8 path (RTX 4090)

heiheiha798:
## Summary

Brings up **Qwen3-VL-8B on Ada RTX 4090 (sm_89)** using the official
`Qwen3-VL-8B-Instruct-FP8` checkpoint. This is the SM89 counterpart to the
SM120/NVFP4 path (#107/#109), but it intentionally uses the checkpoint's
**native block-scaled FP8 language weights** instead of NVFP4 (NVFP4 is a
Blackwell-only path). The vision tower, geometry, scatter, DeepStack and
interleaved-MRoPE are reused from the SM120 Qwen3-VL work; this branch adds
the Ada FP8 language stack and a native sm_89 block-128 GEMM.

The core lever is a **native Ada FP8 e4m3 block-128 scaled GEMM** that reads
the FP8 weight directly (no dequant-to-bf16 scratch), cutting per-linear
weight traffic ~5x while keeping the per-token activation scale. This is the
fix for the SM89 prefill bottleneck (weight-dequant memory bandwidth); the
CUTLASS Ada blockwise route was a dead end at these shapes.

## What's in it

- **Native sm_89 block-128 FP8 GEMM** (`fp8_block128_gemm_mma_sm89.cu`):
  plain Ada `mma.sync.aligned.m16n8k32.row.col.f32.e4m3.e4m3.f32` (no
  `.kind::f8f6f4`, which is sm_120a-only), DeepSeek block-128 scaling folded
  in the mainloop (BLOCK_K pinned to 128). Adapted from the SM120
  `fp8_smallM_handtuned` cp.async pipeline + m16n8k32 tiling. Shape-tuned
  dispatcher over the Qwen3-VL-8B linear shapes.
- **Text frontend** (`qwen3_vl_fp8_sm89.py`): block-scaled M=1 FP8 GEMV
  decode + native block-128 FP8 GEMM prefill; BF16 attention (FA2), BF16
  residual/norms, per-token/per-128 FP8 activation quant fused into
  RMSNorm / residual-add-RMSNorm / SiLU-mul; CUDA-graph decode (per cache
  position). `lm_head` stays **BF16** by default (cuBLASLt); `use_fp8_lm_head`
  is an explicit experimental mode.
- **Multimodal wrapper** (`qwen3_vl_fp8_sm89_multimodal.py`): single-image
  prefill as a captured CUDA-graph replay (vision + S>1 FP8 blockscaled
  language + final norm), lm_head eager after replay; multi-image / video use
  the eager prefill. Decode graphs keyed by `(cache_pos, rope_pos)` so MRoPE
  continuation never reuses a stale rotary position across prompts.
- **FP8 weight loader** (`_qwen3_vl_fp8_weights.py`): keeps the official
  `weight` (e4m3) + `weight_scale_inv` (128x128 block) layout, fuses q/k/v
  and gate/up, optional host-side `lm_head` FP8 quant.
- **Fail-fast & shape guards**: capability `== (8,9)` check, required-symbol
  presence check (rebuild hint), an explicit config check that the checkpoint
  is the official 8B text stack (`36/32/8/128/4096/12288`), and runtime
  `K%128`/`N%128` guards on the new GEMM + block-128 quant entries.
- **Docs / scripts / tests**: `docs/qwen3_vl_fp8_sm89.md`, a smoke benchmark
  (`scripts/smoke_qwen3_vl_fp8_sm89.py`), prefill-GEMM micro-benchmarks, and
  CPU-friendly import/invariant/geometry tests.

Additive on the shared module front: the new GEMM is gated behind
`ENABLE_SM89_BLOCK_FP8_GEMM` (only compiled for `GPU_ARCH=89`); the
block-128 quant helpers are added to the existing quant TU; the SM120 paths
are untouched.

## Performance & correctness (RTX 4090, sm_89)

Environment: NVIDIA GeForce RTX 4090, official `Qwen3-VL-8B-Instruct-FP8`,
workload `FlashRT.png` + prompt `"Describe this image in one sentence."`
(6256 vision patches, 1581 language tokens — same workload as the SM120
report).

```text
S=1581 pixel_shape=(6256, 1536) spans=[(4, 1568)]
set_prompt_warm  median=15.711 ms
vision_only      median=88.990 ms
language_only    median=120.634 ms
prefill          median=206.987 ms  top=785  finite=True
prefill_graph    median=206.214 ms  cos_vs_eager=0.998522  top=785  finite=True
graph_decode     (cache_pos=1581) median=10.681 ms  cos_vs_eager=1.000000  (~94 tok/s)
top_eager=2168   top_graph=2168
```

- Prefill graph eager-vs-replay logits cosine **0.9985**, next-token argmax
  **785** (matches the SM120 report's image token).
- Warm graph decode cosine **1.000000**, decode argmax matches eager
  (`2168`).

The same workload is faster on the SM120/NVFP4 path because it uses Blackwell
NVFP4 language kernels; this branch targets the best Ada-compatible dtype
path (official FP8 weights, BF16 attention/residual/lm_head).

## Build & usage

```bash
cmake --build build -j --target flash_rt_kernels flash_rt_fa2 flash_rt_qwen3_vl_kernels

python examples/qwen3_vl_quickstart.py --arch sm89 \
  --checkpoint /path/to/Qwen3-VL-8B-Instruct-FP8 \
  --image FlashRT.png --prompt "Describe this image in one sentence." \
  --max-seq 2048 --max-prefill-seq 2048

# dev smoke benchmark
python scripts/smoke_qwen3_vl_fp8_sm89.py \
  --checkpoint /path/to/Qwen3-VL-8B-Instruct-FP8 --multimodal --iters 5
```

CPU-friendly tests: `pytest tests/test_qwen3_vl_fp8_sm89.py tests/test_qwen3_vl_smoke.py`.
See `docs/qwen3_vl_fp8_sm89.md` for the full architecture / dtype map / limits.

## Notes (scope)

- Not an NVFP4 implementation — sm_89 uses official FP8 weights by design.
- `lm_head` stays BF16 by default; `use_fp8_lm_head` can match single-step
  logits but multi-token generation can diverge, so it stays experimental.
- Single-image prefill has a CUDA-graph replay path; multi-image / video use
  the eager path.
- Decode graphs are captured per cache position (the FA2 call captures the
  host-side sequence length).
- The frontend is pinned to the official 8B text-stack dims; other Qwen3-VL
  sizes fail fast rather than silently mis-sizing the KV cache.
- The official 8B FP8 weights are too large for CI, so end-to-end numerical
  validation lives in the local smoke script (numbers above); CI covers
  import / loader-invariant / geometry-cache equivalence.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -882,11 +882,12 @@ endif()
 #     from flash_rt import flash_rt_qwen3_vl_kernels as vlk
 #     vlk.rope_neox_qk_bf16(...)
 if(FLASHRT_BUILD_QWEN3_VL)
-  # The Qwen3-VL path depends on the SM120 Qwen3 NVFP4 / FP8 block-128
-  # kernels, so this module only builds for RTX 5090 (sm_120).
-  if(NOT GPU_ARCH STREQUAL "120")
+  # These Qwen3-VL helper kernels are model-agnostic BF16 / FP8 activation
+  # kernels. The SM120 frontend also depends on NVFP4 language kernels, but
+  # the helper module itself can be built for SM89 bring-up as well.
+  if(NOT (GPU_ARCH STREQUAL "89" OR GPU_ARCH STREQUAL "120"))
     message(FATAL_ERROR
-      "FLASHRT_BUILD_QWEN3_VL requires GPU_ARCH=120 (RTX 5090 / sm_120); "
+      "FLASHRT_BUILD_QWEN3_VL requires GPU_ARCH=89 or 120; "
       "got '${GPU_ARCH}'.")
   endif()
   pybind11_add_module(flash_rt_qwen3_vl_kernels
@@ -1083,6 +1084,14 @@ if(GPU_ARCH STREQUAL "120")
     $<TARGET_OBJECTS:sm120_fp8_smallm_obj>)
   target_compile_definitions(flash_rt_kernels PRIVATE
     ENABLE_CUTLASS_SM120_BLOCK_FP8=1)
+endif()
+
+# Native Ada block-scaled FP8 GEMM for official Qwen3-VL FP8 prefill.
+if(GPU_ARCH STREQUAL "89")
+  target_sources(flash_rt_kernels PRIVATE
+    csrc/gemm/fp8_block128_gemm_mma_sm89.cu)
+  target_compile_definitions(flash_rt_kernels PRIVATE
+    ENABLE_SM89_BLOCK_FP8_GEMM=1)
 endif()
 
 # Dedicated M=1 decode GEMV (FP8 + BF16); linked + macro defined wherever the
--- a/csrc/bindings.cpp
+++ b/csrc/bindings.cpp
@@ -10,6 +10,9 @@
 #include "context.h"
 #include "gemm/gemm_runner.h"
 #include "gemm/fp8_block128_gemm.cuh"
+#ifdef ENABLE_SM89_BLOCK_FP8_GEMM
+#include "gemm/fp8_block128_gemm_mma_sm89.cuh"
+#endif
 #ifdef ENABLE_CUTLASS_SM120_BLOCK_FP8
 #include "gemm/cutlass_sm120_block128_fp8_gemm.cuh"
 #include "gemm/fp8_smallM_handtuned_sm120.cuh"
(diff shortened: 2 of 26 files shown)


heiheiha798:
Superseded by a combined 8B + 2B PR (config-driven SM89 path + 2B offline FP8 quantizer). Closing this 8B-only PR; the replacement contains these same commits plus 2B support. New PR link to follow.

PR #111, opened 2026-06-25
Add SM89 Qwen3-VL FP8 path (8B + 2B)

heiheiha798:
## Summary

Brings up **Qwen3-VL on Ada RTX 4090 (sm_89)** with the official block-scaled
**FP8 language weights**, for both the **8B** and the **2B** variants. This is
the SM89 counterpart to the SM120/NVFP4 path (#107/#109), but it intentionally
uses the checkpoint's native FP8 weights instead of NVFP4 (NVFP4 is a
Blackwell-only path). The vision tower, geometry, scatter, DeepStack and
interleaved-MRoPE are reused from the SM120 Qwen3-VL work; this branch adds the
Ada FP8 language stack, a native sm_89 block-128 GEMM, and the config-driven
generalization + offline quantizer that bring up the 2B.

The core lever is a **native Ada FP8 e4m3 block-128 scaled GEMM** that reads
the FP8 weight directly (no dequant-to-bf16 scratch), cutting per-linear weight
traffic ~5x while keeping the per-token activation scale. This is the fix for
the SM89 prefill bottleneck (weight-dequant memory bandwidth); the CUTLASS Ada
blockwise route was a dead end at these shapes.

## What's in it

- **Native sm_89 block-128 FP8 GEMM** (`fp8_block128_gemm_mma_sm89.cu`):
  plain Ada `mma.sync.aligned.m16n8k32.row.col.f32.e4m3.e4m3.f32` (no
  `.kind::f8f6f4`, which is sm_120a-only), DeepSeek block-128 scaling folded
  in the mainloop (BLOCK_K pinned to 128). Adapted from the SM120
  `fp8_smallM_handtuned` cp.async pipeline + m16n8k32 tiling. The dispatcher is
  shape-agnostic (any `N%128==0, K%128==0`), tuned over the Qwen3-VL linear
  shapes.
- **Text frontend** (`qwen3_vl_fp8_sm89.py`): block-scaled M=1 FP8 GEMV decode
  + native block-128 FP8 GEMM prefill; BF16 attention (FA2), BF16
  residual/norms, per-token/per-128 FP8 activation quant fused into RMSNorm /
  residual-add-RMSNorm / SiLU-mul; CUDA-graph decode (per cache position).
  **Now config-driven**: instead of a hard 8B dim guard it enforces the kernel
  constraints (head_dim==128, GEMM dims %128, GQA divisibility) and derives all
  scratch shapes from `config.json`, so the 28-layer/16-head 2B stack runs the
  same kernels with no 8B behavior change.
- **Multimodal wrapper** (`qwen3_vl_fp8_sm89_multimodal.py`): single-image
  prefill as a captured CUDA-graph replay (vision + S>1 FP8 blockscaled
  language + final norm), lm_head eager after replay; multi-image / video use
  the eager prefill. Decode graphs keyed by `(cache_pos, rope_pos)` so MRoPE
  continuation never reuses a stale rotary position across prompts.
- **2B offline quantizer** (`scripts/quantize_qwen3_vl_to_fp8_block128.py`):
  the 2B ships BF16 only, so this writes the official FP8 layout (e4m3
  `weight` + 128x128 `weight_scale_inv` = `amax/448`, multiply-to-dequant),
  keeping norms / `embed_tokens` / the BF16 vision tower unchanged and **not**
  materializing the tied `lm_head`.
- **FP8 weight loader** (`_qwen3_vl_fp8_weights.py`): keeps the official
  `weight` + `weight_scale_inv` layout, fuses q/k/v and gate/up, optional
  host-side `lm_head` FP8 quant; now also handles **single-file** checkpoints
  (no `index.json`) and **tied embeddings** (synthesizes `lm_head` from
  `embed_tokens`) for the 2B.
- **Attention backend** (`attn_backend_qwen3.py`): takes optional
  `num_layers/num_q_heads/num_kv_heads/head_dim` (default to 8B 36/32/8/128)
  and an explicit `device` (was hardcoded `'cuda'` / `current_device()`, wrong
  KV-cache placement under multi-GPU). Additive — the SM120 Qwen3 / Higgs
  callers use the defaults unchanged.
- **Fail-fast & shape guards**: capability `== (8,9)` check, required-symbol
  presence check (rebuild hint), kernel-constraint config check, runtime
  `K%128`/`N%128` guards on the new GEMM + block-128 quant entries, and a
  quantizer guard that the number of quantized linears == `num_layers * 7`.
- **Docs / scripts / tests**: `docs/qwen3_vl_fp8_sm89.md` (8B) +
  `docs/qwen3_vl_fp8_sm89_2b.md` (2B), the smoke benchmark
  (`scripts/smoke_qwen3_vl_fp8_sm89.py`), prefill-GEMM micro-benchmarks, and
  CPU-friendly import/invariant/geometry/quantizer tests.

Additive on the shared-module front: the new GEMM is gated behind
`ENABLE_SM89_BLOCK_FP8_GEMM` (only compiled for `GPU_ARCH=89`); the block-128
quant helpers are added to the existing quant TU; the SM120 paths are
untouched.

## Performance & correctness (RTX 4090, sm_89)

**8B** — official `Qwen3-VL-8B-Instruct-FP8`, workload `FlashRT.png` + prompt
`"Describe this image in one sentence."` (6256 vision patches, 1581 language
tokens — same workload as the SM120 report):

```text
S=1581 pixel_shape=(6256, 1536) spans=[(4, 1568)]
prefill_graph  median=206.2 ms  cos_vs_eager=0.998522  top=785  finite=True
graph_decode   (cache_pos=1581) median=10.681 ms  cos_vs_eager=1.000000  (~94 tok/s)
top_eager=2168  top_graph=2168
```

**2B** — quantized `Qwen3-VL-2B-Instruct-FP8` (FP8 `lm_head`):

```text
text  S=79  decode 2.653 ms = 377 tok/s  logit_cos=0.999426
mm    FlashRT.png S=1581  prefill_graph 99.7 ms  cos_vs_eager=0.9975  top=32
      graph_decode (cache_pos=1581) 2.966 ms = 337 tok/s  cos_vs_eager=1.000000
      generate -> coherent caption of the logo
```

The 2B is ~3.8x faster at decode than the 8B on the same card, as expected
from the ~4x smaller language weights. FP8 `lm_head` is what carries 2B decode
past 350 tok/s (BF16 lm_head: 287 tok/s; same top token at every step). The
same workloads are faster on the SM120/NVFP4 path (Blackwell NVFP4 kernels);
this branch targets the best Ada-compatible dtype path.

## Build & usage

```bash
cmake --build build -j --target flash_rt_kernels flash_rt_fa2 flash_rt_qwen3_vl_kernels

# 8B
python examples/qwen3_vl_quickstart.py --arch sm89 \
  --checkpoint /path/to/Qwen3-VL-8B-Instruct-FP8 \
  --image FlashRT.png --prompt "Describe this image in one sentence."

# 2B: first quantize BF16 -> FP8, then run with --fp8-lm-head
python scripts/quantize_qwen3_vl_to_fp8_block128.py \
  --src /path/to/Qwen3-VL-2B-Instruct --dst /path/to/Qwen3-VL-2B-Instruct-FP8
python examples/qwen3_vl_quickstart.py --arch sm89 --fp8-lm-head \
  --checkpoint /path/to/Qwen3-VL-2B-Instruct-FP8 --image FlashRT.png \
  --prompt "Describe this image in one sentence."
```

CPU-friendly tests: `pytest tests/test_qwen3_vl_fp8_sm89.py tests/test_qwen3_vl_fp8_sm89_2b.py`.
See `docs/qwen3_vl_fp8_sm89.md` / `docs/qwen3_vl_fp8_sm89_2b.md` for the full
architecture / dtype map / limits.

## Notes (scope)

- Not an NVFP4 implementation — sm_89 uses official FP8 weights by design.
- 8B `lm_head` stays BF16 by default; for 2B, `use_fp8_lm_head=True` is
  recommended (it carries decode past 350 tok/s and matches the top token).
- There is no official 2B FP8 release; the 2B FP8 checkpoint is produced
  offline by the repo quantizer.
- Single-image prefill has a CUDA-graph replay path; multi-image / video use
  the eager path. Decode graphs are captured per cache position.
- The official FP8 weights are too large for CI, so end-to-end numerical
  validation lives in the local smoke script (numbers above); CI covers
  import / loader-invariant / geometry / quantizer tests.

_Supersedes #110 (8B-only); this PR is 8B + 2B._


heiheiha798:
### llama.cpp reference comparison

I also checked this PR against a fresh `llama.cpp` SM89 CUDA build as a reference point. This is intentionally kept as a PR note rather than repo documentation, because it is a cross-framework sanity check, not a long-term FlashRT benchmark claim.

Important caveat: `llama.cpp` does not run these HF FP8 safetensors directly. The checkpoints were converted with the current `llama.cpp` FP8 conversion path:

```bash
python convert_hf_to_gguf.py <Qwen3-VL-*-Instruct-FP8> \
  --outtype q8_0 --fp8-as-q8 \
  --outfile <qwen3-vl-*-fp8-as-q8_0.gguf>
```

So the `llama.cpp` rows below are **GGUF Q8_0 converted from FP8**, while FlashRT is running this PR's native official/block-128 FP8 path. This is not a strict native-FP8 apples-to-apples comparison.

Setup summary:

- GPU: single RTX 4090
- FlashRT: this PR, native Qwen3-VL SM89 FP8 path
- llama.cpp: `ggml-org/llama.cpp@683b04cc4`, CUDA build targeting `sm_89`
- Workload: text-only, prefill `S=512`, decode `128` tokens
- llama.cpp command shape: `llama-bench -p 512 -n 128 -ngl 99 -fa on -r 10 -o json`
- FlashRT decode measurement: 128 increasing-position CUDA graph replays after a 512-token prefill; graph capture excluded

| Model | Runtime | Weight format | Prefill 512 | Decode 128 |
|---|---|---|---:|---:|
| Qwen3-VL-2B | FlashRT | native block-128 FP8 | 10.964 ms median, 46.7k tok/s | 350.422 ms median, 365.3 tok/s |
| Qwen3-VL-2B | llama.cpp | GGUF Q8_0 from FP8 | 16.701 ms warm median, 30.7k tok/s | 368.219 ms median, 347.6 tok/s |
| Qwen3-VL-8B | FlashRT | native block-128 FP8 | 41.015 ms median, 12.5k tok/s | 1247.299 ms median, 102.6 tok/s |
| Qwen3-VL-8B | llama.cpp | GGUF Q8_0 from FP8 | 43.171 ms warm median, 11.9k tok/s | 1352.758 ms median, 94.6 tok/s |

Notes:

- The first llama.cpp prefill sample was a cold outlier, so the table uses warm medians from the JSON samples.
- The 8B llama.cpp decode result had some rerun variance; it was in roughly the `95-103 tok/s` bucket. FlashRT's 8B decode replay was stable around `102.6 tok/s`.
- This is a reference validation only. I would not treat it as a permanent cross-framework benchmark unless model format, sampling loop, cache setup, and warmup methodology are standardized.

LiangSu8899:
Thank you very much for this reference comparison!! Overall, the results look as expected, especially the decode phase, which is quite close to my previous comparisons with other adaptations.

PS: Based on past comparison experience, these numbers suggest there's still some room for improvement in the prefill phase.(I would say its like ~1.5-2x more faster) If ultra-low latency is a critical requirement, it might be worth exploring further. I'll also run some profiles on my side to see where we can optimize prefill.

heiheiha798:
Thanks for pointing this out. The PR is still in progress: I am running another independent agent review on the full diff, and I will also take another look at the prefill path to see what can be optimized further.

heiheiha798:
### Prefill optimization (1/2): 2B vision attention head_dim=64 path

Following up on the prefill headroom noted in review. Profiling the **official
FlashRT.png workload** (S=1581, 6256 vision patches) with nsys shows the full
prefill is dominated by two costs, not one — and a language-only profile hides
the second:

| | 2B (`--fp8-lm-head`) | 8B |
|---|---:|---:|
| FP8 block-128 GEMM (language) | 39.5% | 60.0% |
| **Vision-tower attention (FA2)** | **41.8%** | **23.6%** |

The vision tower runs non-causal full attention over all 6256 patches per
block. The 2B vision tower uses `head_dim=64`, but it was being padded into the
FA2 `hd96` bucket, wasting ~33% of the attention work.

**This change** adds a native FA2 `hd64` path:
- `hd64` sm80 forward instantiations (no split-KV — 2B vision never splits, and
  split-KV is the expensive ptxas object).
- A `head_dim<=64` no-split dispatch bucket.
- A `128x64` sm8x non-causal tile, chosen by a new standalone vision-attention
  micro-bench (`benchmarks/sm89_vision_attn`): at S=6256, `128x64 = 1.185 ms` <
  `128x128 = 1.216` < `128x32 = 1.241`.

**Results (RTX 4090, nsys, FlashRT.png):**

| | before | after |
|---|---:|---:|
| 2B vision attention | 43.5 ms | 28.6 ms (**-34%**) |
| **2B total prefill** | 103.9 ms | **89.1 ms (-14.3%)** |
| 8B total prefill | 208.7 ms | 208.6 ms (unchanged) |

8B is intentionally untouched (its vision `head_dim=72` already maps to `hd96`,
which keeps its existing `128x64` tile). Correctness preserved:
`cos_vs_eager=1.000000`, top token unchanged (32), coherent caption.

GEMM-side prefill optimization will follow as a separate change.

heiheiha798:
### Prefill optimization (2/2): FP8 block-128 GEMM access-pattern coalescing

The FP8 block-128 GEMM is the largest single prefill cost (8B 60% / 2B 46% of
GPU kernel time). Using a standalone micro-bench (`benchmarks/sm89_fp8_block128_gemm`,
cold-L2 timing + sampled correctness + NCU wrapper) I iterated NCU-first: profile
the clean baseline kernel, fix the top-ranked bottleneck, re-profile, repeat.

NCU on the baseline (`gate`, M=1581,N=12288,K=4096) ranked the bottlenecks as
memory **access pattern**, not compute or bandwidth (Tensor pipe 35%, DRAM 9%):

**C1 — scale-load coalescing** (NCU top, ~51% est. speedup). The row-strided
scalar `act_scale[row*K128+kb]` loads delivered only 4 of 32 bytes per sector.
Stage the scales in shared memory with a coalesced load, `SCALE_KTILE=8` columns
at a time so the smem footprint is K-independent (~2 KB) and occupancy holds.
(A first attempt staging the whole `[BLOCK_M,K128]` slice regressed `down` +17%
on its 24 KB scale smem — the tiled version is the fix.) NCU: executed
instructions 254M→200M, load sector utilization 4→31.5 of 32 bytes.

**C2 — store-pattern coalescing** (next NCU top, ~47%). The scalar 16-bit BF16
epilogue stores delivered 8 of 32 bytes per sector. The m16n8 layout holds two
adjacent output columns per lane per row, so emit one 32-bit `bfloat162` store
per row. NCU: store sector utilization 8→16 bytes.

**Rejected — occupancy tuning.** NCU still lists occupancy (124 regs → 4 CTAs/SM).
Forcing 5 CTAs/SM drops to 96 regs but spills and runs +4.8% slower: the kernel
is latency-bound (warp-cycles/inst ~14.7, DRAM ~10%), not occupancy-bound.

**Results (RTX 4090, FlashRT.png, nsys, cumulative C1+C2):**

| | before | after |
|---|---:|---:|
| 8B FP8 GEMM kernel | 125.3 ms | 108.7 ms (**-13.2%**) |
| 8B total prefill | 204.4 ms | 187.9 ms |
| 2B total prefill | 89.1 ms | 79.3 ms |

`cos_vs_eager=1.000000`, top token unchanged (8B 785 / 2B 32), decode unchanged.

**Cumulative with the hd64 vision change, vs the pre-optimization baseline:**
2B prefill 103.9 → 79.3 ms (**-23.7%**), 8B prefill 208.7 → 187.9 ms (**-10.0%**).

heiheiha798:
Superseded by #116, which reorganizes the SM89 Qwen3-VL FP8 work into clean, scoped commits aligned with the SM120a path. #116 has been merged. Closing this one.

PR #115, opened 2026-06-26
feat(omnivoice): add OmniVoice TTS FlashRT acceleration backend

shideqin:
## Summary

Adds an OmniVoice TTS acceleration backend to FlashRT, achieving 5.0x throughput
with BF16-quality audio on a single RTX 5060 Ti (SM120).

Key results: 100 ms latency, RTF 0.032, 2.0 GB VRAM, mel-cosine 0.9961 vs PyTorch BF16.

## Performance

| Metric | Value |
|---|---|
| Latency (ns=32, gs=2.0) | 100 ms |
| Speedup vs PyTorch | 5.0x |
| RTF | 0.032 |
| VRAM | 2.0 GB (70% of PyTorch) |
| Mel-cosine (hybrid) | 0.9961 |
| Mel-cosine (BF16 baseline) | 0.9971 |
| Max abs error (hybrid) | 0.0009 |
| Token match rate (hybrid) | 99.7% |

Strategy: BF16 CFG on step 1 establishes token structure (17/816 tokens,
2.9% mutated). FP4 noCFG on steps 2-32 fills remaining tokens at 5x speed.
Perceptually identical to full BF16.

### CMake + Bindings
- `FLASHRT_ENABLE_OMNIVOICE` option (default OFF, requires GPU_ARCH=120)
- Gates `flash_rt_omnivoice` module behind the flag
- `flash_rt_kernels.so` contains zero OmniVoice-specific symbols

### CUDA Kernels (`csrc/kernels/omnivoice/`)
- `omnivoice_cfg_logsoftmax_bf16` — fused CFG combine (3x log_softmax to 1 kernel)
- `omnivoice_qk_norm_rope_bf16` — warp-per-head Q/K RMSNorm+RoPE
  (qkv_split + norm + rope: 3 launches to 1)

All kernels are namespaced with `omnivoice_` prefix, gated by
`FLASHRT_ENABLE_OMNIVOICE=ON`.

### Python Pipeline (`flash_rt/models/omnivoice/`)
- `FlashRTLlm` — FP4 W4A4 forward with CUDA Graph, FA2, fused kernels
- `FlashRTLlmBF16` — BF16 forward, zero quantization overhead
- `inject()` / `free_encoder()` / `eject()` public API
- `_check_kernels()` — validates required kernel symbols before model loading
- OmniVoice package import is lazy (inside `_optimize_maskgit()`)

### Serving + Examples
- `serving/omnivoice_agent/server.py` — FastAPI server with POST /predictions
- `examples/omnivoice_quickstart.py` — CLI quickstart

### Tests (`tests/test_omnivoice_smoke.py`)
- 11 tests, skip cleanly when compiled modules are absent (flag OFF)
- No hardcoded paths, follows repo conventions

### Documentation (`docs/PERFORMANCE_OMNIVOICE.md`)
- RTF, mel-cosine, max abs error specs
- VRAM breakdown by component
- Build instructions and API reference


LiangSu8899:
Thanks a lot for the substantial OmniVoice work here. The direction is valuable: the PR adds a model-specific TTS acceleration path, keeps `flash_rt_omnivoice` as an optional CMake target, and the standalone OmniVoice module itself builds cleanly with `FLASHRT_ENABLE_OMNIVOICE=ON` on SM120.

Before merge, I think we should tighten a few integration/compliance points so this lands cleanly and stays maintainable long-term:

## Merge-before checklist

- [ ] **Keep OmniVoice as a fully optional dependency**
  - `flash_rt.models.omnivoice` currently imports `omnivoice.models.omnivoice` at module import time.
  - Please move the OmniVoice package import into the runtime path that actually needs it, for example inside `inject()` / `_optimize_maskgit()`.
  - A user without the external OmniVoice package installed should still be able to import `flash_rt` and ideally import `flash_rt.models.omnivoice` enough to see a clear runtime error from `inject()`.

- [ ] **Fix the smoke tests to match the real API surface**
  - The tests currently check `flash_rt.inject`, `flash_rt.api.FlashRTLlm`, `api._has_cfg_kernel`, and `api._check_kernels`, but the implementation exposes these under `flash_rt.models.omnivoice`.
  - Please either update the tests to import from `flash_rt.models.omnivoice`, or explicitly add a public API export if that is intentional. My preference is to keep the root `flash_rt` API clean and use `flash_rt.models.omnivoice`.
  - Default-build tests should skip cleanly when compiled `.so` modules are absent, not fail with `ModuleNotFoundError`.

- [ ] **Remove environment-specific test path injection**
  - `tests/test_omnivoice_smoke.py` currently defaults to `/data/omnivoice_flashrt_bundle`.
  - Please remove the hard-coded path, or only add a bundle path when `OMNIVOICE_BUNDLE` is explicitly set by the user/test environment.

- [ ] **Align docs with the actual import path**
  - The docs currently show `from flash_rt import inject, free_encoder`.
  - If the API remains model-scoped, please change this to `from flash_rt.models.omnivoice import inject, free_encoder`.

- [ ] **Clean up model-specific kernel naming**
  - The new kernels are model/shape specialized, but some names look generic: `fused_qk_norm_rope`, `cfg_combine`, `maskgit_sample`, `maskgit_select`.
  - For long-term maintainability, please give model-specific kernels an explicit model prefix in file names, C++ symbols, and pybind symbols, for example:
    - `omnivoice_qwen3_qk_norm_rope_bf16`
    - `omnivoice_cfg_logsoftmax_bf16`
    - `omnivoice_maskgit_sample_row_bf16`
    - `omnivoice_maskgit_select_topk_bf16`
  - Also prefer a model-specific folder such as `csrc/kernels/omnivoice/` so the common kernel namespace does not become ambiguous.
  - Please avoid experimental suffixes such as `v4` in public pybind names unless there is an intentional versioned API contract.

- [ ] **Either wire `maskgit_*` kernels into the runtime path or mark/remove them**
  - `maskgit_sample_row_bf16` and `maskgit_select_topk_bf16` are exported and tested, and the docs list them as part of the accelerated module.
  - However, the current Python path still uses the PyTorch `_gumbel_sample` / `torch.topk` / scatter path in `_sample_step()`.
  - Please either connect these kernels to the real hot path, or remove them from required symbols/docs for this PR and leave them for a follow-up.

- [ ] **Keep default builds clean and non-invasive**
  - The standalone `flash_rt_omnivoice` target being gated by `FLASHRT_ENABLE_OMNIVOICE=ON` is good.
  - Please make sure any OmniVoice-only kernel that does not need to live in `flash_rt_kernels` is also gated or clearly justified as shared/common.

## What I verified locally

- `git diff --check origin/main...HEAD`: passed.
- Python compile for the new OmniVoice files: passed.
- Default SM120 `flash_rt_kernels` build with `FLASHRT_ENABLE_OMNIVOICE=OFF`: passed.
- Gated `flash_rt_omnivoice` target with `FLASHRT_ENABLE_OMNIVOICE=ON`: passed, and the three exported symbols are importable.
- Current `tests/test_omnivoice_smoke.py` does not pass in a clean default environment because it checks the wrong API layer and does not skip missing compiled modules cleanly.

Overall, this is promising work, but I would prefer to merge after the optional dependency boundary, tests/docs, and kernel namespace are cleaned up. That will keep the main FlashRT surface clean while still letting OmniVoice land as a well-isolated acceleration backend.

shideqin:
@LiangSu8899 All checklist items addressed and pushed:

1. OmniVoice import is now lazy (inside `_optimize_maskgit()`), `import flash_rt` works without omnivoice package
2. Smoke tests updated to `flash_rt.models.omnivoice`, skip cleanly when .so absent
3. Hardcoded bundle path removed, no `OMNIVOICE_BUNDLE` env var
4. Docs updated with correct import paths
5. Kernels moved to `csrc/kernels/omnivoice/`, renamed with `omnivoice_` prefix, no `v4` suffix
6. `maskgit_sample`/`maskgit_select` fully removed
7. `flash_rt_kernels.so` has zero OmniVoice symbols

Build verified: all 3 .so modules compile clean, 11/11 smoke tests pass.

PTAL when you have a moment, thanks!

LiangSu8899 (approved):

PR #116, opened 2026-06-26
Add SM89 Qwen3-VL FP8 path (8B + 2B)

heiheiha798:
## Summary

Brings up **Qwen3-VL on Ada RTX 4090 (sm_89)** with the checkpoint's native
block-scaled **FP8 language weights**, for both the **8B** and **2B** variants.
This is the SM89 counterpart to the SM120/NVFP4 Qwen3-VL path: the vision tower,
geometry, image scatter and MRoPE are reused from that work; this branch adds
the Ada FP8 language stack, a hand-written sm_89 block-128 GEMM/GEMV, and the
offline quantizer + config-driven generalization that bring up the 2B.

This is a clean re-cut of the earlier prototype branch onto current `main`,
reorganized into reviewable commits with the file scope aligned to the SM120
Qwen3-VL pattern (dedicated `flash_rt_qwen3_vl_kernels` module; no central
`bindings.cpp` churn; iteration/benchmark harnesses kept out of the PR).

## Why hand-written (no TMA on Ada)

The SM120 Qwen3-VL GEMM runs through the CUTLASS
`KernelTmaWarpSpecializedBlockwise` collective — TMA bulk-copy + a
warp-specialized producer/consumer pipeline with an auto-selected deep
(3-4 stage) software pipeline that hides load latency without spending
registers or occupancy on staging.

**Ada (sm_89) has no TMA and no warp-specialized CUTLASS collective** (those
builders are Sm90+/Sm120 only), so that design does not port:

- **GEMM** is a hand-written `cp.async` + `ldmatrix.x4` + block-128-scaling MMA
  kernel (`mma.sync...m16n8k32...e4m3.e4m3.f32`, no `.kind::f8f6f4`). With
  `cp.async` instead of TMA, each extra pipeline stage costs a full smem buffer;
  at the tuned 64x64 tile the kernel is already register-limited to 4 CTA/SM, so
  a 2-stage pipeline halves occupancy and regresses. It runs single-stage and is
  latency-bound. The obvious memory-pattern wins (load/store coalescing,
  `ldmatrix`) are applied; CUTLASS's own Ada blockwise example is slower than
  this kernel at these shapes, so it is not used.
- **Attention** uses vendored FlashAttention-2 (Tri Dao). The only addition is
  an `hdim64` instantiation for the 2B ViT (avoids a 64->96 pad); the FA2 kernel
  body is upstream and unmodified.

So the SM89 path matches the SM120 design *intent* (block-128 FP8, fused
act/norm-quant, ldmatrix MMA) but cannot reuse the TMA/warp-specialized
machinery — the main reason its prefill is slower than the Blackwell NVFP4 path.

## Commits

1. `csrc: add SM89 native FP8 block-128 GEMM kernel`
2. `csrc: add SM89 block-128 M=1 GEMV for Qwen3-VL decode`
3. `csrc: add SM89 Qwen3-VL helper kernels (qk-norm-rope, fp8 block act/norm)`
4. `build: bind SM89 Qwen3-VL kernels into flash_rt_qwen3_vl_kernels`
5. `fa2: add hdim64 instantiation for Qwen3-VL 2B vision attention`
6. `frontend: SM89 Qwen3-VL official-FP8 path (8B + 2B)`
7. `tools: SM89 FP8 block-128 quantizer + smoke/CPU tests`
8. `docs: SM89 Qwen3-VL FP8 (8B + 2B); note no-TMA optimization limits`

## Module layout

SM89-specific kernels (block-128 GEMM/GEMV, fused act/norm-quant, QK
norm-rope-kvwrite, the ViT bf16 cuBLASLt matmul) build into the dedicated
`flash_rt_qwen3_vl_kernels` module (`FLASHRT_BUILD_QWEN3_VL=ON`, now `GPU_ARCH`
89 or 120), keeping `flash_rt_kernels.so` small. The frontend resolves names
from the dedicated module first, then the generic one — the same dual-module
split the SM120 ViT path already uses.

## Build

```bash
cmake --build build --target flash_rt_kernels flash_rt_fa2 flash_rt_qwen3_vl_kernels -j
```

## Verification

Built clean for sm_89 (both modules). End-to-end 2B multimodal smoke
(`scripts/smoke_qwen3_vl_fp8_sm89.py --multimodal --fp8-lm-head`, FlashRT.png):
`cos_vs_eager=1.000000`, top token matches eager, prefill ~80 ms, decode
`cos_vs_eager=1.000000`, coherent generation. CPU/CI tests cover the config
guard, the block-128 quant round-trip, and the expected quant-linear count
without the multi-GB checkpoint or CUDA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,8 +133,8 @@ endif()
 #                              pi0.5 / groot need bf16. Default "fp16;bf16".
 set(FA2_ARCH_NATIVE_ONLY OFF CACHE BOOL
     "Only build FA2 for the detected arch (no sm_80/PTX fallback). ~66% faster build.")
-set(FA2_HDIMS "96;128;256" CACHE STRING
-    "Semicolon-separated FA2 head_dim instantiations to build. Shipped models use 96 and 256.")
+set(FA2_HDIMS "64;96;128;256" CACHE STRING
+    "Semicolon-separated FA2 head_dim instantiations to build. Qwen3-VL 2B vision uses 64; other shipped models use 96 and 256.")
 set(FA2_DTYPES "fp16;bf16" CACHE STRING
     "Semicolon-separated FA2 dtype instantiations to build. Pi0 uses fp16; pi0.5/groot use bf16.")
 
@@ -704,8 +704,8 @@ if(ENABLE_FA2)
   set(FA2_BUILT_HDIMS "")
   set(FA2_BUILT_DTYPES "")
   foreach(_hd IN LISTS FA2_HDIMS)
-    if(NOT _hd MATCHES "^(96|128|256)$")
-      message(FATAL_ERROR "FA2_HDIMS entry '${_hd}' not supported. Valid: 96, 128, 256.")
+    if(NOT _hd MATCHES "^(64|96|128|256)$")
+      message(FATAL_ERROR "FA2_HDIMS entry '${_hd}' not supported. Valid: 64, 96, 128, 256.")
     endif()
     list(APPEND FA2_BUILT_HDIMS ${_hd})
     foreach(_dt IN LISTS FA2_DTYPES)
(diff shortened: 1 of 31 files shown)


LiangSu8899:
Very cool!

LiangSu8899 (approved):

PR #120, opened 2026-06-27
build: add opt-in FLASHRT_SLIM_BUILD to trim VLA-deployment kernel compile surface

heiheiha798:
## Summary

Adds an opt-in `FLASHRT_SLIM_BUILD` CMake option (default **OFF**) that drops
model/architecture-specific CUDA translation units the target hardware/model
path does not need, reducing `flash_rt_kernels` compile time for VLA
deployments on weaker build machines.

This is **compile-time surface reduction only**. It does not change kernel math,
launch parameters, thresholds, dtype conversions, dispatch policy, graph-capture
behavior, Python API semantics, or runtime fallback. The default build keeps the
exact broad, compatibility-oriented kernel set it has today; slimming is fully
opt-in.

## How it works

Each gated group is guarded by its own internal `FLASHRT_HAVE_*` compile
definition so the corresponding pybind bindings in `csrc/bindings.cpp` drop in
**lockstep** with their sources — a gated source whose binding stayed compiled
would fail to link. Architecture-keyed groups reuse the existing arch macros
(`ENABLE_NVFP4`, `GPU_ARCH`) so real NVFP4/SM120 builds always retain them, slim
or not.

Gated groups (slim mode):
- **Motus VAE FP8** (5 TUs) — only the cleanly Motus-isolated quantize kernels.
  Shared FP8 helpers used by other models (GROOT AdaLN, MelBand GeLU-FFN) stay
  always-compiled.
- **Qwen3.6 / linear-attention** (10 TUs) — Qwen3.6-family-only kernels. Shared
  helpers reused by Qwen3-dense / Higgs / Cosmos3 stay always-compiled. Neutral
  helpers (`bf16_matmul_bf16`, `embedding_lookup_bf16`) are never gated.
- **SM120 / NVFP4-named** (7 TUs) — reuse `ENABLE_NVFP4` / `GPU_ARCH` so SM120
  and NVFP4 builds keep them in both modes.

## Tooling

- `scripts/build_inventory.py` reports the directly-compiled TUs per pybind
  target from the configured build dir (generator-agnostic, via
  `DependInfo.cmake`), categorizes the `flash_rt_kernels` set, and accounts for
  object-library TUs linked via `$<TARGET_OBJECTS:...>` so the reported compile
  surface is not understated.
- `tests/test_build_inventory.py` and `tests/test_generic_kernel_bindings.py`
  are build-mode aware: they read `FLASHRT_SLIM_BUILD` from `CMakeCache.txt` and
  assert the matching surface and binding contract. Compat asserts legacy +
  neutral bindings present; slim asserts the gated legacy bindings are absent
  while neutral bindings remain.

## Validation

**SM89 (RTX 4090):**
- Default `flash_rt_kernels`: 55 direct (+2 object) TUs.
- Slim `flash_rt_kernels`: 33 direct (+2 object) TUs (−22 direct).
- `flash_rt_qwen3_vl_kernels` (9) and `flash_rt_fa2` (1+20) unchanged.
- Binding tests pass in both modes; legacy Qwen3.6 bindings present in default,
  absent in slim; neutral bindings present in both.

**Thor (compute capability 11.0, `GPU_ARCH=110`):**
- Default `flash_rt_kernels`: 59 direct (+8 object) TUs.
- Slim `flash_rt_kernels`: 37 direct (+8 object) TUs (−22 direct).
- Slim removes exactly the same 22 direct TUs as SM89 and does **not** remove
  architecture-specific object libraries Thor requires.
- Binding checks match the intended contract in both modes.

Both architectures confirm: slim removes exactly the intended 22 direct TUs,
default compatibility builds retain the broad historical bindings, and no
architecture-specific object libraries needed at runtime are dropped.

## Scope notes

- Two mis-named-but-shared files per group (Motus and Qwen3.6) are intentionally
  kept always-compiled in this PR; neutralizing their ownership is deferred to a
  follow-up to avoid moving kernel bodies here.
- Default-ON / smarter hardware-profile defaults are intentionally deferred
  until more hardware/model paths are validated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1028,6 +1028,19 @@ if(GPU_ARCH STREQUAL "120" OR GPU_ARCH STREQUAL "121" OR
 endif()
 
 # ── Main module ──
+# flash_rt_kernels is the shared core pybind module. It currently compiles a
+# broad superset of kernels, including several model/architecture-specific
+# translation units that not every deployment needs (Qwen3.6/linear-attention,
+# SM120/NVFP4-named, Motus/video FP8-history). For VLA deployments on weaker
+# machines this is avoidable build latency.
+#
+# An incremental cleanup gates those model/arch-specific TUs behind explicit
+# CMake options so a slim build compiles only what a given hardware/model path
+# needs. This is COMPILE-TIME surface reduction only: it does not change kernel
+# math, launch params, dtype/dispatch policy, graph capture, or runtime
+# fallback. Default builds stay compatibility-oriented (broad set) until the
+# slim path is validated across SM89/SM120/Thor. See scripts/build_inventory.py
+# and tests/test_build_inventory.py for the measured per-target surface.
 pybind11_add_module(flash_rt_kernels
   csrc/gemm/gemm_runner.cu
   csrc/kernels/norm.cu
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,6 +168,24 @@ option(FLASHRT_ENABLE_MELBAND_ROFORMER
 # Enable explicitly on a Blackwell (sm_120) build.
 option(FLASHRT_ENABLE_OMNIVOICE
        "Build OmniVoice TTS RTX SM120 fused kernels in flash_rt_omnivoice" OFF)
+# ── Slim build for VLA deployments ──
+# OFF by default: the standard build keeps the broad, compatibility-oriented
+# kernel set it has today. Turn ON (-DFLASHRT_SLIM_BUILD=ON) to drop
+# model/architecture-specific translation units that a given hardware/model
+# path does not need, reducing flash_rt_kernels compile time on weaker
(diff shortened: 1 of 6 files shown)


LiangSu8899:
I added a small public docs patch on top of this PR:

- `docs/INSTALL.md` now documents `FLASHRT_SLIM_BUILD=ON` as an optional VLA-oriented slim build.
- It states that the default build remains the broad compatibility surface.
- It lists the gated groups and warns not to use slim mode for Qwen3.6 / Nex-N2, Motus FP8/VAE, or non-VLA NVFP4 conversion flows.
- It also calls out that this is a first step toward explicit build profiles, not yet a general `vla` / `llm` / `vlm` / `tts` / `video` profile system.

Local checks for the docs patch:

- `git diff --check`: passed
- privacy scan on `docs/INSTALL.md`: no local paths/usernames/debug traces found

LiangSu8899 (approved):

PR #130, opened 2026-07-03
qwen36: DFlash speculative decoding path for Thor (drafter window, constant-time rollback, opt-in acceptance modes, serving host)

LiangSu8899:
## Summary

Brings the DFlash block-diffusion drafter path up on Jetson AGX Thor (SM110) and closes the two structural gaps that kept it below the MTP chain. On structured robot-plan prompts the DFlash path now decodes at **48.9 tok/s vs 33.7 tok/s** for the FP8-KV MTP reference (AL 4.57 vs 2.87), with **byte-identical greedy output** on every measured prompt.

New doc: `docs/qwen36_dflash.md` (checkpoint, usage, window knobs, measured numbers, benchmark caveats).

## Commits

1. **fix(qwen36_thor): match parent `_long_tq_effective_k` signature** — the shared frontend grew an optional `max_new_tokens` parameter and passes it positionally; the Thor override kept the two-argument form, so every long-context generate on Thor raised `TypeError`. All 11 Thor overrides were audited for signature drift; this was the only one.

2. **feat(qwen36): DFlash speculative decoding on Thor** — the DFlash loop hard-coded the BF16-staged verify and per-position prompt walk (RTX-shaped). Split both behind arch hooks with defaults preserving the RTX flow byte-for-byte; Thor routes the verify through the FP8-KV wrapper (same as the long-ctx spec verify), prefills through the chunked FP8 path, and guarantees the FP8 KV cache exists after drafter load on short-context constructions.

3. **perf(qwen36): constant-time partial-accept rollback** — the loop re-advanced committed rows with a second full main-model forward per cycle; on weight-read-bound Thor this doubled per-cycle traffic (152.8 ms/cycle). Thor now grows the per-step state checkpoints to the DFlash verify q_seq and rolls back with two `gpu_copy` calls (same pattern as the long-context MTP loop): **152.8 -> 92.9 ms/cycle**. Both layer types delegate K <= `_K_save_max` to the parent K-row so committed rows and any recovery recompute stay on one kernel family — committing rows from one family while recovering with another surfaces their occasional rounding disagreements as greedy divergence (observed and root-caused during bring-up).

4. **feat(qwen36): per-token drafter window with prompt-tail seeding** — the drafter window held one fc-projected feature per spec cycle, i.e. entries ~AL tokens apart read as consecutive positions. The per-token window keeps one entry per committed token (appended in bulk outside the drafter graph; the graph itself is read-only over the window and needs no capture snapshot), and the Thor prefill seeds it from the prompt tail. Default on for Thor, opt-in on RTX; env-gated (`FLASHRT_QWEN36_DFLASH_PERTOKEN`, `..._WINDOW`, `..._WINDOW_SEED`).

5. **docs(qwen36)** — usage doc, env table, README news entry.

6. **fix(qwen36): commit per-token window rows before the tap shuffle** (review follow-up) — the window append ran after the end-of-cycle `taps[:, 0]` shuffle, so each cycle's first committed feature entered the window as a duplicate of the last accepted row. The append is now `_dflash_window_commit(N)`, called before the shuffle. All performance claims re-measured with the corrected window and updated everywhere (the buggy window happened to inflate the JSON-plan number; the honest figure is 48.9 tok/s).

7. **tests: structural coverage without checkpoints** (review follow-up) — `tests/test_qwen36_dflash_structural.py`: window-commit row order + copy semantics, a source-order guard on the generate loop, missing-drafter fail-fast, public drafter-init delegation, and Thor per-token env routing (default / opt-out / window override). 8 tests, CPU-only, no model files.

8. **feat(qwen36): public `init_dflash_drafter`** (review follow-up) — public wrapper for the drafter load; docs no longer instruct users to call an underscored method.

9. **feat(serving): `serving/qwen36_dflash_agent`** — stateless OpenAI-compatible host for the DFlash path per the serving-layer contract (policy above the frontend, no exec/ verbs): /v1/chat/completions, /v1/models, /health, batch-1 serialized, accept-length telemetry in the response. Long-running sessions stay on `serving/qwen36_agent` (MTP path); the README states the split.

## Measured (Thor SM110, steady state, 64/256-token delta, greedy)

| prompt | MTP AL / tok/s | DFlash AL / tok/s | parity |
|---|---:|---:|---|
| robot task -> JSON plan | 2.87 / 33.7 | **4.57 / 48.9** | PASS |
| robot navigation plan | 2.59 / 30.5 | 3.25 / 34.8 | PASS |
| prose explanation | 2.43 / 28.5 | 3.00 / 31.7 | PASS |
| repeated-sentence stress text | 3.84 / 36.8 | 2.13 / 22.9 | PASS |

Parity is against the FP8-KV MTP route (`FLASHRT_QWEN36_LONG_CTX_ROUTE_MIN_SEQ=0` for short prompts) — the BF16 short route stores KV in a different format, so token-exact comparison across formats is not meaningful. The repeated-sentence row is a benchmark pathology: the seeded window is fed degenerate context; `FLASHRT_QWEN36_DFLASH_WINDOW_SEED=0` recovers AL 4.27 there.

## Regression evidence

- MTP spec path unchanged: 3.84 AL / 37.0 tok/s / 103.9 ms cycle at ctx=128 before and after (same-process A/B).
- RTX default flow: all new behavior is behind arch hooks whose defaults reproduce the previous code path byte-for-byte, and the per-token window is default-off outside Thor.
- Greedy parity: 3 consecutive 256-token DFlash runs bit-identical, all matching the MTP reference.
- Per-step checkpoint growth reallocates the save buffers at drafter load and drops K-row graphs captured against the old allocations, so pre-existing MTP graphs cannot hold stale pointers.

## Review follow-ups addressed

- Blocking: per-token window tap-order bug — fixed, re-measured, numbers corrected in docs/README/PR.
- Blocking: no test coverage — structural test suite added (runs without checkpoints or GPU).
- Minor: underscored public entry — `init_dflash_drafter` added and documented.
- Minor: host-sync in the decode loop — documented in `docs/qwen36_dflash.md` (one `argmin().item()` per cycle, ~10 us against an ~86 ms verify, included in all measurements; device-side accept noted as follow-up).

## Risk

R3 (touches the shared DFlash loop and Thor dispatch). No kernel or build changes; `csrc/` untouched. Thor-validated on hardware; RTX validated by default-path preservation (hooks default to prior behavior).


LiangSu8899:
TBD: Add a speculative decoding abstraction for multi-backend support.

LiangSu8899:
Follow-up: decode-ceiling round (borrowed from the TRT-LLM/vLLM/SGLang playbook, measured on Thor).

Cycle breakdown first (CUDA events, steady state): verify 84.1 ms / drafter 8.1 ms / everything else 1.2 ms — so scheduler/host-overlap techniques from the serving engines have nothing to buy here; the whole budget is the verify.

Three additions, defaults chosen by the greedy-parity red line:

1. **Relaxed thinking-phase acceptance** (opt-in, `FLASHRT_QWEN36_DFLASH_RELAXED_THINKING=1`) — the TensorRT-LLM MTP policy: inside `<think>`, accept drafts in the verify top-k within a logit margin; strict matching resumes at `</think>` so the visible answer stays exact-verified. Thinking-enabled robot-plan prompt: AL 3.78 -> 5.42, **40.4 -> 57.7 tok/s (+43%)**. CPU unit tests cover the acceptance math (top-k membership, margin cutoff, strict-after-close).

2. **Chunk-saves verify kernels** (opt-in, `FLASHRT_QWEN36_THOR_LIN_CHUNK_SAVES=1`) — chunk-scan variants that emit the per-step rollback checkpoints in one pass: verify 84.1 -> 79.5 ms (~+5% decode). Kept off by default: they move the S=8..16 verify onto the Thor kernel family while the parity reference runs the parent family, and the transcript drifts (measured). The K<=7 production MTP dispatch is untouched either way.

3. Docs for both knobs, with the tradeoffs stated.

Default configuration is re-verified byte-identical to the FP8-KV MTP reference (parity PASS, numbers unchanged: 93.5 ms cycle / AL 2.13 on the parity prompt). 11 structural tests pass without checkpoints or GPU.

PR #146, opened 2026-07-12
Add sm_120 INT4 (E0M3) W4A4 GEMV kernel and OMMA unlock tool

LiangSu8899:
## What

A native **INT4 (E0M3)** W4A4 GEMV for **sm_120** (RTX 50-series), plus the
build-time tool that unlocks it, a synthetic-data standalone test, and a doc.

On sm_120 the single block-scaled 4-bit tensor-core instruction
`mma.sync.kind::mxf4nvf4.block_scale.scale_vec::4X.m16n8k64` (SASS
`OMMA.SF.16864`) carries the operand **element format** in bits **78 (A)** and
**79 (B)** of the 128-bit encoding: `0 = e2m1` (NVFP4), `1 = e0m3` (uniform
INT4, codebook −7…+7). `ptxas` only accepts `e2m1`, so the kernels are compiled
as NVFP4 and post-processed to INT4. The block-scale path (UE4M3 per-16 SF +
fp32 global scale) is unchanged, so INT4 reuses the whole NVFP4 quant/swizzle
machinery.

This element-format finding was documented publicly by the **Ling Team**
(@im0qianqian): https://zhuanlan.zhihu.com/p/2059376150565089368 — this PR
reproduces and productizes it for FlashRT.

## Why INT4 over E2M1

E2M1's non-uniform grid has an inherent *shrinkage bias* (asymmetric rounding
bins pull values toward zero); a uniform 4-bit grid removes it, so per-block
scaling and Hadamard rotation translate into real quality. Bit-faithful
simulation of the shipped two-level quantizer (identical UE4M3 per-16 SF + fp32
global scale; `amax/(7·448)` for INT4 vs `amax/(6·448)` for E2M1) on real
weights — INT4 lowers weight quant error (1 − cosine) on **37/37** sampled
tensors:

| tensor family | mean quant-error reduction (best INT4 vs E2M1) |
|---|---|
| diffusion DiT (attn + FFN) | −15 % |
| 3D-conv VAE (residual convs) | −18 % |
| dense transformer (attn + MLP + lm_head) | −15 % |

Uniform-grid + per-16 Hadamard wins on every tensor; the same rotation under
E2M1 never helps — matching the shrinkage-bias explanation (Ling UFP4 report).

## Throughput — identical to NVFP4

Same instruction, so INT4 runs at the NVFP4 rate. Issue-rate microbenchmark,
register-resident MMA (no HBM traffic), **RTX 5090, CUDA 13.0**, median of 5:

| operand formats | throughput |
|---|---|
| NVFP4 `e2m1 × e2m1` | 2026.6 TFLOPS |
| INT4 `e0m3 × e0m3` | 2026.8 – 2027.1 TFLOPS |
| mixed `e0m3 × e2m1` | 2027.9 TFLOPS |

*Metric: instruction issue rate (2·M·N·K per MMA), no memory traffic — this is
the tensor-core rate, not an end-to-end GEMM.* A prebuilt HF kernel running the
same microbenchmark: `huggingface.co/kernels/flashrt/int4-blackwell`.

## Contents

- `csrc/kernels/int4_w4a4_mma_sm120.{cu,cuh}` — M=1 full-N W4A4 GEMV (bf16 out),
  INT4 block-scale quantizer, `amax/(7·448)` global-scale reduction, a runtime
  **codebook canary**, and `extern "C"` wrappers for a GGML/llama.cpp backend.
  Same fragment layout and cp.async double-buffering as the NVFP4 twin
  `fp4_w4a4_mma_sm120`.
- `tools/patch_int4_omma_sm120.py` — flips OMMA bits 78/79 in device functions
  whose mangled name contains a marker (`int4_`); works on a bare `.cubin` or in
  place on a host ELF (`.so`/executable) with an uncompressed `.nv_fatbin`.
  `--operands b` patches only the weight operand (mixed W-INT4 × A-E2M1).
- `csrc/kernels/test_int4_w4a4_standalone.cu` — synthetic-data correctness (cos
  vs a plain-C++ INT4 reference) + issue-rate microbench.
- `docs/int4_blackwell.md`.

## Compiled E2M1, RUN as E0M3

`ptxas` rejects any element type but `e2m1` here, so **the patch step is part of
linking**. An unpatched binary decodes E2M1 on INT4 data (wrong results,
undetectable by the launcher), so module load must gate on
`int4_w4a4_sm120_codebook_canary() == 0` (fail-fast). The bits are undocumented
and invisible to `nvdisasm`/`cuobjdump` (a patched instruction still
disassembles as `E2M1.E2M1`); re-verify the encoding on any CUDA-toolkit bump —
the canary is the authoritative check.

## How to run

sm_120 GPU + CUDA 13.0+. No checkpoints — synthetic data.

```bash
nvcc -std=c++17 -O3 -gencode arch=compute_120a,code=sm_120a \
  csrc/kernels/int4_w4a4_mma_sm120.cu \
  csrc/kernels/test_int4_w4a4_standalone.cu -o <out>/test_int4
python tools/patch_int4_omma_sm120.py <out>/test_int4
<out>/test_int4
```

Observed on RTX 5090 / CUDA 13.0 (build `arch=compute_120a`; plain
`compute_120` rejects `mxf4nvf4`):

```text
[canary] 0  (E0M3 decode OK)
[gemv] N=4096 K=4096  cos(kernel, INT4-ref) = 0.999431  (expect > 0.999)
[bench] issue-rate: 2012.4 TFLOPS  (INT4 x INT4, 680 blocks x 8 warps)
[result] PASS
```

`cos(kernel, INT4-ref)` compares the kernel against a C++ implementation of the
identical INT4 two-level recipe — validates SF swizzle addressing + fragment
layout (residual is bf16-output rounding vs fp32 reference).

## Scope / isolation

**Experimental, additive, isolated.** No `CMakeLists.txt` / pybind changes: this
is a runtime primitive validated by a standalone `nvcc` test, exactly like the
existing `csrc/**/test_*_standalone.cu`. The default build path is unchanged.
CMake/pybind wiring and the end-to-end model gate (token-exact decode) land in a
follow-up PR. The natural first consumer is a 4-bit GGML backend on Blackwell
(hardware dequant in the tensor core at the NVFP4 rate) — see `docs/int4_blackwell.md`
for the GGUF Q4_0/Q4_K mapping notes.

## Credits & references

- Element-format bits documented by the **Ling Team**, @im0qianqian:
  https://zhuanlan.zhihu.com/p/2059376150565089368
- Ling Team, *Rethinking Shrinkage Bias in LLM FP4 Pretraining: Geometric
  Origin, Systemic Impact, and UFP4 Recipe* (arXiv).

### DIFF
--- a/csrc/kernels/int4_w4a4_mma_sm120.cu
+++ b/csrc/kernels/int4_w4a4_mma_sm120.cu
@@ -1,0 +1,469 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// Tensor-core INT4 (E0M3) W4A4 GEMV + quantizer for sm_120. See the
+// header (int4_w4a4_mma_sm120.cuh) for the format contract and the
+// mandatory OMMA bit-patch post-build step. Fragment layout notes
+// live in fp4_w4a4_mma_sm120.cu — this TU uses the identical layout,
+// expressed as raw PTX so the file stays free of CUTLASS includes
+// (self-contained for the llama.cpp backend build).
+
+#include "int4_w4a4_mma_sm120.cuh"
+
+#include <cuda_bf16.h>
+#include <cuda_runtime.h>
+#include <cstdint>
+
+namespace flash_rt {
+namespace gemm {
+
+namespace {
+
+// ── Block-scaled MMA (compiled e2m1, patched to E0M3) ─────────────
+//
+// Same operand order as cute SM120_16x8x64_TN_VS<..., VS=16>::fma.
+// scale_vec::4X — each lane's sfa/sfb u32 carries the 4 UE4M3 bytes
+// for its fragment rows/cols across the 4 K-groups of this K-tile.
+
+__device__ __forceinline__ void int4_mma_16x8x64(
+    float& d0, float& d1, float& d2, float& d3,
+    uint32_t a0, uint32_t a1, uint32_t a2, uint32_t a3,
+    uint32_t b0, uint32_t b1,
+    float c0, float c1, float c2, float c3,
+    uint32_t sfa, uint32_t sfb) {
+  constexpr uint16_t bid = 0, tid = 0;
+  asm volatile(
+      "mma.sync.aligned.kind::mxf4nvf4.block_scale.scale_vec::4X.m16n8k64"
+      ".row.col.f32.e2m1.e2m1.f32.ue4m3 "
+      "{%0,%1,%2,%3},{%4,%5,%6,%7},{%8,%9},{%10,%11,%12,%13},"
+      "{%14},{%15,%16},{%17},{%18,%19};\n"
+      : "=f"(d0), "=f"(d1), "=f"(d2), "=f"(d3)
+      : "r"(a0), "r"(a1), "r"(a2), "r"(a3), "r"(b0), "r"(b1),
+        "f"(c0), "f"(c1), "f"(c2), "f"(c3),
+        "r"(sfa), "h"(bid), "h"(tid),
+        "r"(sfb), "h"(bid), "h"(tid));
+}
+
+// ── UE4M3 helpers (same semantics as the NVFP4 quantizers) ────────
+
+__device__ __forceinline__ uint8_t fp32_to_ue4m3_ceil(float v) {
+  if (v <= 0.0f) return 0;
+  if (v > 240.0f) return 0xFE;
+  uint32_t bits = __float_as_uint(v);
+  int float_exp = ((bits >> 23) & 0xFF) - 127;
+  uint32_t frac = bits & 0x7FFFFF;
+  int ue_exp = float_exp + 7;
+  if (ue_exp <= 0) {
+    float scaled = v * 512.0f;
+    int m = (int)ceilf(scaled);
+    if (m > 7) return (1 << 3) | 0;
+    if (m < 1) m = 1;
+    return (uint8_t)m;
+  }
+  if (ue_exp >= 15) return 0xFE;
+  int m = (int)(frac >> 20);
+  if (frac & 0xFFFFF) m++;
+  if (m >= 8) { m = 0; ue_exp++; }
+  if (ue_exp >= 15) return 0xFE;
+  return (uint8_t)((ue_exp << 3) | m);
+}
+
+__device__ __forceinline__ float ue4m3_to_fp32(uint8_t v) {
+  int e = (v >> 3) & 0xF;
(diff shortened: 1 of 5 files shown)


LiangSu8899:
Thanks for the review — all three points were right. Pushed `0082e88`.

**1. `--verify` was not a real gate (main issue).** Correct, and it was the exact wrong thing to be soft about. The root cause is that static verification is *asymmetric*: an unpatched binary disassembles fine, but once bits 78/79 are set `cuobjdump` can no longer decode the instruction as `OMMA.SF` (I confirmed both failure shapes — it either errors out or silently emits no OMMA line), so a patched binary can't be re-read statically at all. New semantics:

- `--verify --expect e2m1` — reliable **pre-patch gate**: exit 0 iff every site is still E2M1, nonzero otherwise (including "no sites found").
- `--verify --expect int4` — exit **1** on an unpatched binary (NOT patched), exit **2** (inconclusive) on a patched one. It never returns 0 by guessing.
- patch mode now exits **3** when it flips zero instructions (already patched / no marker match), so a build step fails loudly instead of silently shipping.

Verified exit codes end to end:

```
pre-patch  --expect e2m1  -> exit 0  OK: all 120 site(s) are 'e2m1'
pre-patch  --expect int4  -> exit 1  FAIL: 120/120 site(s) are not 'int4'
patch                     -> exit 0  patched 120 OMMA.SF instruction(s)
post-patch --expect int4  -> exit 2  INCONCLUSIVE ... run int4_codebook_canary()
re-patch (already)        -> exit 3  nothing patched ... may already be patched
runtime canary            -> 0       E0M3 decode OK
```

The runtime canary `int4_w4a4_sm120_codebook_canary()` stays the authoritative check for a patched binary — the docstring and doc now say this explicitly.

**2. Doc inconsistency (line 102).** You're right, and my original claim was just wrong: a patched instruction does **not** still disassemble as `E2M1.E2M1` — it becomes unreadable. Corrected the toolchain note to describe the asymmetry and point at the canary.

**3. Non-blockers, both taken.** Host wrappers now return a nonzero code on `cudaGetLastError()` after launch (was always returning 0). And the `extern "C"` symbols are prefixed `flashrt_int4_sm120_*` to avoid colliding with a host app when linked alongside GGML/llama.cpp.

Isolation is unchanged: still no `CMakeLists.txt`/pybind/Python-runtime edits; default build unaffected. The CMake/pybind wiring + token-exact decode E2E gate remain a follow-up PR.

Issue #143, opened 2026-07-11
Track Jetson-PI multi-backend provider integration


heiheiha798:
## Purpose

This issue tracks review of the FlashRT multi-backend model-provider integration corresponding to Jetson-PI.

- FlashRT implementation branch: https://github.com/heiheiha798/FlashRT/tree/jetson-pi-link-check
- FlashRT branch tip: `28c3adf6c152609cb9379572b9916a53368b0420`
- Companion Jetson-PI branch: `flashrt-migration-merge` in the private Jetson-PI repository
- Companion private Jetson-PI PR: https://github.com/diantoudedianshan/Jetson-PI/pull/1
- Jetson-PI reviewer/owner: @diantoudedianshan

The FlashRT branch is hosted on the contributor fork because the current contributor account has read-only access to the official FlashRT repository.

## Scope

The branch introduces Jetson-PI as a provider-owned llama.cpp/GGML model provider while preserving the existing FlashRT CUDA Graph path. It includes Pi0/Pi0.5, text LLM, and multimodal LLM paths; staged runtime ports; repeatable decode; memory-domain and read-only DLPack integration; production CMake packaging; and tested CPU/CUDA/Vulkan/SYCL/OpenCL capability behavior.

## Review status

Development and available-backend validation are complete. This work is pending review by @diantoudedianshan. The Jetson-PI owner will decide how the private implementation should be opened and integrated; this issue only tracks the completed development branch and review handoff.

The companion Jetson-PI PR is private and is accessible only to authorized repository members.

LiangSu8899:
Super COOL!!!

LiangSu8899:
@LiangSu8899 Verify interface completeness on branch after integration 
https://github.com/flashrt-project/FlashRT/pull/144
https://github.com/LiangSu8899/FlashRT-Nexus/pull/12

heiheiha798:
Update: the FlashRT-side integration is now up as a public PR — **#148** (`feat/jetson-pi-provider`, rebased on `origin/main`).

Backend reference moved from the original private fork to the **public** repository [`PKU-SEC-Lab/Jetson-PI-Edge`](https://github.com/PKU-SEC-Lab/Jetson-PI-Edge) (`Jetson-PI-flashrt` branch). The switch is **zero code changes** on the FlashRT side — integration is path-parameterized via `JETSON_PI_ROOT` / `JetsonPI_ROOT`, no hardcoded git URLs.

[...]

PR #148, opened 2026-07-15
feat: integrate Jetson-PI as Pi0/LLM/MLLM provider with verified staged plans

heiheiha798:
## Summary

Integrate [Jetson-PI](https://github.com/PKU-SEC-Lab/Jetson-PI-Edge) as an
optional provider-owned llama.cpp/GGML backend in FlashRT, exposing Pi0, LLM,
and MLLM through the `frt_model_runtime_v1` generic stage-plan ABI. The existing
FlashRT CUDA Graph path is unchanged.

> **Updated onto current main.** This branch carries the v1-generic refactor originally reviewed in #157 (commit `048cd21`). The parallel v2 ABI was removed in favor of the merged additive v1 generic stage-plan extension from #156.

## What Changed

- Added the Jetson-PI provider, C ABI bridge, packaging, Python frontends, and
  Pi0/LLM/MLLM provider tests.
- LLM and MLLM publish provider-owned `reset -> prefill -> decode` plans.
- Pi0 publishes `context -> action` only after both the source feature contract
  and the loaded Jetson-PI DSO runtime capability confirm a real split.
- The capability stays in the existing 32-bit reserved engine word, avoiding a
  vtable size change.
- Pi0 state uses a bucket-variable dimension. The Python frontend forwards the
  finite policy state without action-width padding; the backend enforces the
  selected model's state bound.
- Pi0 configuration rejects `n_views` outside `[1, 3]` in Python, JSON runtime,
  direct runtime creation, and the engine factory.
- Pi0 runtime-open releases the transferred owned engine reference when a
  successful factory result has a safely readable release hook but an otherwise
  incomplete vtable.
- Real-model tests require explicit checkpoint-specific action dimensions. The
  verified `pi0_base` checkpoint is 50x32; the tested PI0.5 Libero checkpoint
  is 10x32.
- Added old-vtable, capability-gating, view-limit, state-contract,
  invalidation, one-shot, and real-model parity coverage.

## Companion Backend PR

[PKU-SEC-Lab/Jetson-PI-Edge#1](https://github.com/PKU-SEC-Lab/Jetson-PI-Edge/pull/1)
is the merged baseline (`ab4dd79`). The coordinated follow-up
[PKU-SEC-Lab/Jetson-PI-Edge#2](https://github.com/PKU-SEC-Lab/Jetson-PI-Edge/pull/2)
at reviewed head `3a269c7` provides the required runtime contract:

1. model-specific legacy Pi0 and PI0.5 prepare/decode paths with owned pending
   context storage;
2. PI0.5 prompt-state serialization and legacy Pi0 tensor-state semantics;
3. pointer/count, finite-state, model-width, view-count, and exact combined
   batch-capacity validation;
4. scoped PI0.5 combined-batch ownership on all error paths;
5. a runtime capability query that distinguishes the fixed DSO from historical
   same-SONAME builds;
6. preserved legacy Pi0 vs PI0.5 GGUF conversion metadata.

The current FlashRT provider build requires that runtime capability API. An old
same-SONAME DSO without the symbol fails to load explicitly; it is not silently
treated as a real split implementation.

## Validation

Final cross-repository review used FlashRT `fd3f177` and Jetson-PI-Edge
`3a269c7`.

- FlashRT CPU provider CTest: **12/12 passed**.
- FlashRT CUDA provider CTest: **12/12 passed**.
- Python state-contract tests: **2/2 passed**.
- Malformed successful factory results release their owned engine reference;
  CPU and CUDA provider regressions pass.
- Jetson-PI no-weight capability/view/prompt tests and upstream mtmd C API tests
  pass on CPU and CUDA.
- Real legacy `pi0_base` CUDA state/split passes at 50x32; real PI0.5 Libero
  CUDA state/split passes at 10x32. Both reject oversized combined batches and
  retain bit-identical whole vs split output (`max abs diff = 0`).
- Real legacy FlashRT CUDA wrapper/direct parity passes with `max abs diff = 0`.
- Real PI0.5 FlashRT CUDA state formatting/sensitivity and split/backend parity pass across state boundary regions, invalidation, and one-shot consumption. This gate does not claim action parity against the reference PI0.5 policy.
- FlashRT Python CUDA end-to-end returns finite nonzero `[50, 32]` legacy
  actions; `context/action` is bit-identical to `predict()`.
- The provider DSO has a strong undefined reference to
  `jetson_pi_pi0_capabilities`; `ldd -r` against the historical DSO reports the
  missing symbol as expected.
- The provider DSO surface remains limited to `frt_model_runtime_open_v1`.

## Out of Scope

- Deduplicating the three C++ runtime wrappers, the hand-written JSON parser,
  Python staged-decode frontends, and LLM/MLLM engine classes.
- Jetson-PI's existing batched-VIT-to-sequential fallback.
- ConfidenceHead / DSpark-style refinement.
- Internal per-checkpoint state normalization; callers provide the normalized
  policy state expected by Jetson-PI.

Related issue: #143. Refactor base: #157. v1 capability extension: #156.
Companion backend follow-up: PKU-SEC-Lab/Jetson-PI-Edge#2.


LiangSu8899:
Hi @heiheiha798, first of all, thank you very much for putting this integration together. The provider boundary, Jetson-PI packaging work, Pi0/LLM/MLLM engine coverage, and parity tests gave us a solid implementation to build on. I kept your commits and authorship in the replacement branch.

One shared-runtime detail changed while this PR was open: model-runtime v1 now has the additive generic stage-plan extension from #156. With that available, keeping the parallel v2 from this branch would leave us with two public authorities for ports, stages, identity, and lifecycle. That would be difficult for both providers and hosts to maintain long term.

I prepared #157 on top of your work to try a narrower version of the same integration:

- every provider returns `frt_model_runtime_v1`;
- `context -> action` and `reset -> prefill -> decode` are published through the generic OPAQUE plan;
- `step()` remains the whole-model fallback;
- outputs use the normal STAGED `get_output` path;
- Jetson-PI/GGML state stays completely provider-owned;
- the provider no longer changes `runtime/`, `exec/`, or shared kernels.

I reran the CPU/provider suites, installed-DSO checks, and a cross-repository Nexus Pi0 provider contract E2E. I have not replaced your real-model GPU evidence with a new claim; that still needs target hardware and weights for a fresh release run.

Could you please take a look at #157 when you have time, especially the selected-plan mapping and whether the v1 wrappers still preserve the execution semantics you intended? I would really appreciate your opinion, and I am happy to adjust the refactor if I misunderstood any part of the Jetson-PI side. I will leave this PR open while we discuss it rather than closing your contribution unilaterally.

LiangSu8899:
Hi @heiheiha798, thanks again for the integration. We ran the public Jetson-PI path with a real converted PI0.5 checkpoint and then checked both the revision recorded by this PR (`5de3f9e`) and the current `Jetson-PI-flashrt` branch. Could you help confirm whether the following two points are intentional or are missing wiring?

1. The server path has `build_pi05_openpi_prompt()` and puts the discretized state into `Task: ..., State: ...;\nAction:`, but `jetson_pi_pi0_context()` always zero-pads state and calls `llama_set_pi0_state()`. The PI0.5 graph does not consume that separate state tensor. In our run, changing raw state from zeros to ones produced bit-identical actions, while putting the same state into the PI0.5-formatted prompt changed the result. Is the narrow C API supposed to format PI0.5 state itself, or is the caller expected to pass an already formatted prompt? The current FlashRT usage/API appears to promise raw prompt plus state.

2. The repository already has `mtmd_helper_prepare_chunks_pi0_for_model()` and `mtmd_helper_decode_pi0()`, plus `pending_context`, but the narrow `context()` currently calls the whole `mtmd_helper_eval_chunks_pi0()`, stores the final action in `action_cache`, and `action()` then copies it. We measured the split as roughly 264 ms in `context()` and 0.15 ms in `action()`. Was `context -> action` intended to be a real encode/decode boundary, or only a lifecycle-compatible result handoff?

If these are intentional contracts, we can adjust the provider schema and documentation accordingly. If they are implementation gaps, we can keep the provider on a single `infer` stage until the narrow API exposes the intended behavior. Would appreciate your guidance before we change the integration semantics.

heiheiha798:
Thanks for the careful real-checkpoint validation. You are right on both points.

1. The caller is not supposed to pre-format a PI0.5 prompt. The narrow API was intended to accept the raw task text and state, with Jetson-PI owning model-specific prompt/state serialization. The current `llama_set_pi0_state()` plus zero-padding path reflects the legacy Pi0 tensor-state assumption; it is not the correct PI0.5 contract when the policy consumes the discretized proprioception values in the text prompt.

For PI0.5, we should expose the policy's actual state shape separately from `action_dim`, accept that state on the API boundary, and build the same `Task: ..., State: ...;\nAction:` representation internally. We should also add reference-policy parity coverage that changes state while holding images/task fixed, so this cannot regress silently.

2. `context -> action` was intended to be a real pending-context/decode split, not merely a lifecycle-compatible result handoff. The current implementation does not satisfy that intent: `context()` runs the full `mtmd_helper_eval_chunks_pi0()` path and caches the final action, while `action()` only returns the cached buffer. The existing `pending_context`, `mtmd_helper_prepare_chunks_pi0_for_model()`, and `mtmd_helper_decode_pi0()` path is the direction that needs to be completed.

So I agree with your proposed integration treatment: until the narrow API exposes and validates that real boundary, please publish Pi0 as a single `infer` stage in #157. The `context -> action` selected plan, documentation, and staged tests should not be advertised for Pi0 yet. Once the Jetson-PI API is corrected and has reference-policy/state parity plus a one-shot pending-context test, we can re-enable that plan in a follow-up.

The v1 generic-plan refactor itself still looks like the right direction. Thanks for preserving the provider boundary and for catching the distinction between wrapper/direct parity and PI0.5 reference-policy parity.

heiheiha798:
Final cross-PR review is complete against FlashRT `77a0949` and companion
Jetson-PI-Edge `afa0d37`.

The issues raised in the earlier review are now addressed:

- FlashRT forwards the non-empty policy state at its actual width instead of
  padding PI0.5 state to `action_dim`.
- The backend documents and enforces the model-specific state contract;
  PI0.5 NULL state now matches an explicit eight-value zero state.
- The real pending-context/decode split has no cached-action fast path and its
  one-shot behavior is covered.
- Standalone backend CTest discovery works from the repository root without
  registering the backend test when embedded by FlashRT.
- The dynamic CHECK formatting bug, dead cache state, stale comments, and PR
  documentation have been corrected.

Validation completed on the updated heads:

- FlashRT CPU and CUDA provider CTest: 12/12 passed in each build.
- Backend CPU and CUDA root CTest: 1/1 passed in each build.
- Real PI0.5 backend state/split tests passed on CPU and RTX 4090, including
  state boundary sensitivity, NULL/explicit-zero equivalence, oversized-state
  rejection, one-shot behavior, and split-vs-whole max absolute difference 0.
- FlashRT CUDA state parity passed for all boundary regions.
- FlashRT CUDA wrapper/direct parity passed for 320 outputs with max absolute
  difference 0.
- Python end-to-end inference returned finite nonzero `[10, 32]` actions, and
  the pure-Python state-width regression passed.

I found no remaining implementation blockers across the pair. For complete
PI0.5 state correctness, merge
[Jetson-PI-Edge#1](https://github.com/PKU-SEC-Lab/Jetson-PI-Edge/pull/1)
before or together with this PR. FlashRT intentionally keeps Pi0 on a single
`infer` stage until that backend change is consumed; the selected
`context -> action` plan can be enabled in a follow-up.

heiheiha798:
Cross-PR review correction after the independent xhigh pass: my earlier final
comment was too broad and is superseded by this one.

The follow-up review found concrete blockers, now fixed at FlashRT `b9b042a`
and Jetson-PI-Edge `4ced592`:

- FlashRT publishes a bucket-variable Pi0 state dimension instead of claiming
  an exact `[8]` ABI shape.
- The backend rejects state pointer/count mismatches, model-overwidth state,
  and NaN/Inf before prompt discretization.
- PI0.5 prompt formatting is shared by production and deterministic tests;
  exact openpi bins, short state, and NULL-state text are asserted.
- The standalone test follows `LLAMA_BUILD_TESTS`, excludes CMake.js, and does
  not leave stale CTest registrations after an ON-to-OFF reconfigure.
- Newly added real-model state tests require explicit backend/model-kind
  selection. Single-use file helpers were inlined.
- The test now has an explicit legacy Pi0 branch for short-state zero padding,
  full-width state, state effect, split equivalence, and width rejection.

Revalidation is green: FlashRT CPU and CUDA provider CTest are 12/12; backend
CPU and CUDA selected CTest are 1/1; CPU and RTX 4090 PI0.5 state/split tests
pass; FlashRT CUDA state parity passes all four bins; wrapper/direct parity is
exact (`320` values, max absolute difference `0`); Python CUDA E2E returns
finite nonzero `[10, 32]` actions.

Final decision: I see no remaining implementation blocker for the exercised
PI0.5 cross-PR path after these fixes. I am not extending that conclusion to a
legacy Pi0 real-model run: the branch is compiled and test-covered, but this
environment has no accessible legacy checkpoint, so that remains an explicit
validation gap rather than an inferred pass.

heiheiha798:
Final cross-PR review update (supersedes my previous status comments).

Reviewed pair:

- FlashRT: `ce97f9b`
- Jetson-PI-Edge follow-up #2: `25533aa` (intentionally remains Draft)

The remaining cross-repository gaps are now closed:

- FlashRT publishes Pi0 `context -> action` only when the complete engine
  vtable and the explicit real-split capability are both present. Older
  backends and callback-only cached-result shims remain on `infer`.
- The capability uses the existing reserved engine word, and the precise
  `run_stage` size gate preserves the v1 allocation boundary.
- Jetson-PI routes detected legacy Pi0 through the legacy evaluator, retains an
  owned prepared batch (including copied VIT embeddings), and keeps PI0.5 on
  its model-specific path.
- Pending contexts are invalidated on replacement and consumed once. GPU
  encoded-KV storage is not freed while decoder graphs reference it; a prefix
  size change resets those graphs before reallocation.
- GGUF conversion no longer forces legacy Pi0 to `pi0.pi05=true`, and the
  real-model test asserts the detected family before inference.

Final validation on the reviewed heads:

- FlashRT provider CTest: 12/12 CPU and 12/12 CUDA.
- Jetson-PI selected state/split plus upstream mtmd CTest: 2/2 CPU and 2/2 CUDA.
- Real PI0.5 CUDA state/split and FlashRT wrapper/direct/Python paths pass;
  whole vs split and FlashRT vs direct have `max abs diff = 0`.
- `lerobot/pi0_base` was downloaded through `hf-mirror`, converted to a legacy
  GGUF (`pi0.pi05=false`), and run on CPU and CUDA. CUDA backend state/split
  passed 5/5 repeated runs; whole vs split has `max abs diff = 0`.
- Legacy FlashRT CUDA wrapper/direct parity is exact for all 1600 outputs, and
  Python `predict()` is bit-identical to `context()`/`action()`.
- The canonical three-view legacy layout was also exercised on CUDA, covering
  the third owned VIT embedding buffer; output was finite/nonzero `[50, 32]`
  and whole vs split remained exact.

I found no remaining implementation blocker in the reviewed pair. Follow-up #2
is deliberately still Draft and should not be marked ready or merged
automatically. This FlashRT change remains safe before #2 lands because absence
of the explicit capability selects the single `infer` plan.

heiheiha798:
Final cross-PR review update after the post-review fixes. This supersedes my
earlier final comment.

Reviewed heads: FlashRT `9236e83`, Jetson-PI-Edge `3a269c7`.

All findings from the final independent pass are now addressed:

- the loaded Jetson-PI DSO is queried for the real context/action capability;
  an older same-SONAME DSO cannot be mistaken for a capable runtime;
- Pi0 rejects more than three views at every public FlashRT entry point and in
  the backend before model loading;
- legacy Pi0 and PI0.5 preflight exact text/image token usage before allocating
  or writing the fixed combined batch;
- PI0.5 combined-batch ownership covers all early error paths;
- real-model tests require checkpoint-specific action shapes, with verified
  `pi0_base` 50x32 and tested PI0.5 Libero 10x32 values.

Final validation:

- FlashRT provider CTest: 12/12 on CPU and 12/12 on CUDA;
- Python state-contract tests: 2/2;
- real legacy Pi0 and PI0.5 backend CUDA state/split suites pass, including
  oversized-prompt rejection and split-vs-whole `max abs diff = 0`;
- FlashRT legacy CUDA wrapper/direct parity and PI0.5 CUDA state parity pass;
- Python legacy CUDA end-to-end returns finite nonzero `[50, 32]` actions and
  staged execution is bit-identical to `predict()`;
- the provider's strong capability-symbol dependency fails explicitly against
  the historical DSO, as intended.

Decision: I find no remaining implementation blocker in the reviewed cross-PR
path. Jetson-PI-Edge #2 is intentionally still Draft and should remain so until
the coordinated merge/update decision is made.

heiheiha798:
Final cross-PR review update after the post-review fixes. This supersedes my
earlier final comment.

Reviewed heads: FlashRT `fd3f177`, Jetson-PI-Edge `3a269c7`.

All findings from the final independent pass are now addressed:

- the loaded Jetson-PI DSO is queried for the real context/action capability;
  an older same-SONAME DSO cannot be mistaken for a capable runtime;
- Pi0 rejects more than three views at every public FlashRT entry point and in
  the backend before model loading;
- legacy Pi0 and PI0.5 preflight exact text/image token usage before allocating
  or writing the fixed combined batch;
- PI0.5 combined-batch ownership covers all early error paths;
- real-model tests require checkpoint-specific action shapes, with verified
  `pi0_base` 50x32 and tested PI0.5 Libero 10x32 values.
- malformed successful factory results release their transferred owned engine
  reference whenever the declared struct safely exposes `self` and `release`.

Final validation:

- FlashRT provider CTest: 12/12 on CPU and 12/12 on CUDA;
- Python state-contract tests: 2/2;
- malformed-factory ownership regressions pass on CPU and CUDA;
- real legacy Pi0 and PI0.5 backend CUDA state/split suites pass, including
  oversized-prompt rejection and split-vs-whole `max abs diff = 0`;
- FlashRT legacy CUDA wrapper/direct parity and PI0.5 CUDA state parity pass;
- Python legacy CUDA end-to-end returns finite nonzero `[50, 32]` actions and
  staged execution is bit-identical to `predict()`;
- the provider's strong capability-symbol dependency fails explicitly against
  the historical DSO, as intended.

Decision: I find no remaining implementation blocker in the reviewed cross-PR
path. The coordinated review is complete, and Jetson-PI-Edge #2 can now be
marked Ready for maintainer review.

LiangSu8899 (approved):
Final accession review completed after updating onto current main. Default isolation, provider-only contracts, Python state contract, real CPU/CUDA backend parity evidence, capability fail-closed behavior, and the runtime/exec/csrc zero-diff boundary are all satisfied. The companion Jetson-PI backend PR remains the merge-order dependency; no FlashRT implementation blocker remains.

PR #157, opened 2026-07-20
Refactor Jetson-PI provider onto model runtime v1

LiangSu8899:
## Summary

This keeps the Jetson-PI Pi0/LLM/MLLM integration contributed in #148, but rebases its public boundary onto the existing `frt_model_runtime_v1` capability model.

The provider remains optional and owns all GGML/llama.cpp semantics. FlashRT runtime core, exec, shared kernels, and existing graph producers are unchanged.

## Why this refactor

#148 originally introduced a parallel model-runtime v2 to represent provider-owned stages and memory. Since then, the v1 ABI gained an additive, size-probed generic stage-plan extension. Maintaining both runtime families would give producers and hosts two competing authorities for the same ports, stages, identity, and lifecycle.

This version therefore:

- returns `frt_model_runtime_v1` from every provider path;
- publishes one selected generic OPAQUE plan through `query_extension`;
- retains `step()` as whole-model compatibility sugar;
- uses ordinary STAGED `get_output` instead of a provider-specific memory token;
- keeps provider state and executor references opaque to FlashRT and Nexus;
- removes all provider changes from `runtime/`, `exec/`, and `csrc/`.

## Selected plans

- Pi0 fallback: `infer`
- Pi0 staged: `context -> action`
- LLM/MLLM fallback: `infer`
- LLM/MLLM staged: `reset -> prefill -> decode`

Executor references intentionally differ from stage indices and are validated by the provider contract tests.

## Build and packaging

- The provider and Jetson-PI integration remain model-level opt-in features and are disabled by default.
- Provider-only CPU builds do not require CUDA or FlashRT exec.
- The installed provider exports only `frt_model_runtime_open_v1`.
- Provider dependencies use a relocatable `$ORIGIN` lookup within the deployment prefix.

## Validation

- [x] Default provider-off build does not create provider targets
- [x] Provider-only CPU contract suite: 3/3
- [x] Jetson-PI CPU suite: 11/11
- [x] Pi0/LLM/MLLM selected-plan and whole-step tests
- [x] Direct-provider parity fixtures remain byte/token exact
- [x] Installed DSO dependency closure and handle-local `dlsym`
- [x] Export allowlist contains only `frt_model_runtime_open_v1`
- [x] Nexus generic adopter cross-repository Pi0 provider contract E2E
- [x] SM110 CUDA build and real PI0.5 GGUF checkpoint open/inference
- [x] Real two-view output is finite, nonzero, repeat bit-exact, and wrapper/direct-stage bit-exact
- [x] `runtime/`, `exec/`, and `csrc/` have zero diff from current main
- [x] Public docs and artifacts contain no development paths or environment records
- [ ] PI0.5 reference-policy state/prompt/token/action parity
- [ ] Installed-package Jetson-PI dependency route reproduced on the SM110 target

The real-checkpoint run at `048cd21b36517c79cebaa93c5c46c90d60a5deee` measured a 62.358 s open and a 266.857 ms steady two-view p50 for the converted F16 GGUF path. These are qualification measurements, not production performance claims.

The run also found two model-level blockers that the wrapper/direct parity test cannot detect:

1. The current Pi0 adapter exposes `state[action_dim]`, but the tested PI0.5 policy has an eight-value proprioceptive state whose normalized and discretized values belong in the formatted prompt. Changing the raw staged state produced bit-exact actions, while correctly formatting zero versus one state changed actions (`cos=0.990631708`, `max_abs=1.555626512`).
2. The current Jetson-PI `context()` computes and caches the final action, while `action()` only copies it (`264.385 ms` versus `0.145 ms`). Until the backend exposes a real pending-context/action boundary, the provider should publish one `infer` stage rather than imply a computational split.

The ABI/lifecycle refactor remains independently reviewable, but this draft is not yet a PI0.5 release candidate. Direct Jetson-PI versus FlashRT-wrapper parity proves that the adapter preserves the backend result; it does not prove equivalence to the PI0.5 reference policy.

## Relationship to #148

This is a reviewable replacement built directly on the original contribution and preserves its commit authorship. I am leaving #148 open while we discuss the boundary with the contributor.


LiangSu8899:
Superseded by the contributor-owned #148, which now carries the reviewed v1-generic refactor, the coordinated backend fixes, and the current-main integration update. Continuing review and merge qualification on #148 preserves the external contribution as the canonical accession PR.

PR #164, opened 2026-08-05
pi05 thor update: NVFP4 encoder/decoder tier + FA4 attention

LiangSu8899:
## Summary

Brings the Pi0.5 Thor NVFP4 work up to the current production configuration. The FP8 path is unchanged; everything new sits behind `use_fp4=True` and its sub-flags.

Same-session A/B against the FP8 path on Jetson AGX Thor (locked clocks, MAXN, LIBERO fixtures, 10 denoise steps, medians over 100 iterations after 20 warmup), measured with the committed harness:

| Views | FP8 | NVFP4 + FA4 | Speedup | Action cosine (worst sample) |
|---:|---:|---:|---:|---:|
| 3 | 49.02 ms | **31.74 ms** (32 Hz) | **1.54x** | 0.999736 (0.999435) |
| 2 | 38.70 ms | **27.17 ms** (37 Hz) | **1.42x** | 0.999720 (0.999157) |
| 1 | 32.92 ms | **23.01 ms** (43 Hz) | **1.43x** | 0.993474 (0.971119) † |

Each row is one same-session A/B; rows are separate sessions, and the FP8 baseline drifts a few ms between sessions on this hardware, so only within-row ratios are meaningful. Both children run FA4, so the speedup isolates NVFP4 rather than the attention backend.

A second session re-ran all three view counts through the harness default `--construct load_model`, which builds both children with the public `load_model()`. Every cosine came out identical at every view count and the FP4 latencies landed within 0.15 ms (23.07 / 27.29 / 31.89 ms), so the two construction paths demonstrably produce the same configuration. Both tables and the artifact hashes are in `docs/pi05_thor_decoder_fp4_e2e.md`.

The 2- and 3-view runs clear every gate the harness enforces (raw and action cosine, per-sample floors, and the latency targets). † At 1 view the per-sample gate does not pass; `docs/pi05_thor_decoder_fp4_e2e.md` characterizes this as a flow-matching bifurcation rather than a kernel defect — with single-view input the velocity field itself is near a decision boundary on some samples — and records an encoder-FP8 + rotated-INT4-decoder configuration that does clear the gates at 1 view. The README row is marked accordingly.

## What is in the port

- **NVFP4 decoder** — all four projections (QKV, O, Gate+Up, Down) as block-scaled NVFP4 with narrow-N v10 tiles, plus a fused GeGLU epilogue that folds the combiner into the Gate+Up GEMM.
- **NVFP4 encoder** — 17-layer FFN preset with AWQ, the P1 split-GU path with a half-width fused GeGLU epilogue, and NVFP4 attention output projections.
- **NVFP4 SigLIP FFN** — extended to all 27 layers by an AWQ rescue, with vectorized single-pass LayerNorms feeding it.
- **E0M3 (uniform-INT4) decoder** — weights via a runtime instruction descriptor, optional INT4 activations with per-16 Hadamard rotation, all behind explicit flags and defaulted off.
- **FA4 attention** for SigLIP and the encoder, and vectorized QKV split-RoPE / FP4 quantize kernels.
- **A scale-factor zero-initialization fix** (`csrc`/`fp4_utils`): the tile-interleaved SF layout rounds K up to 64-element atoms, and the quantize kernels only write real coordinates, so allocation garbage in the padding could decode as UE4M3 NaN and poison a block-scaled accumulator. This one is not Pi0.5-specific — any NVFP4 caller with a non-multiple-of-64 K was exposed.

## Public API

`use_fp4_decoder=True` selects the complete tier the table above is measured with, resolving the sub-flags left at `None` the same way `awq_alpha` already keyed off it:

```python
model = flash_rt.load_model(
    checkpoint, config="pi05", framework="torch", hardware="thor",
    num_views=3, use_fp4=True, use_fp4_decoder=True, use_fa4=True)
```

That resolves to `use_fp4_encoder_attn=True`, `use_fp4_siglip_ffn=True`, `encoder_p1_combiner="epilogue_hw"` and `awq_alpha=0.8`. `use_fp4=True` alone keeps the earlier encoder-only preset (`False`, `False`, `"lut_native"`, `0.5`). Any sub-flag passed explicitly still overrides the preset.

`use_fa4` is an attention-backend choice and is deliberately independent of `use_fp4`: both the FP8 frontend and the NVFP4 subclass accept it, and the FP8 baseline above runs it too.

## Behavior changes to call out

- `use_fp4=True` now resolves its encoder preset to `tuple(range(17))` instead of `range(18)`, and `awq_alpha` defaults to `None` (0.8 with the FP4 decoder, 0.5 otherwise) instead of a flat 0.5.
- `use_fp4_decoder=True` additionally resolves `encoder_p1_combiner` to `"epilogue_hw"` (was `"lut_native"`) and turns on the NVFP4 encoder attention output and SigLIP FFN, which is what makes the published latency reachable from `load_model()`.
- New `load_model` options: `use_fp4_decoder`, `use_fp4_encoder_attn`, `use_fp4_siglip_ffn`, `use_fa4`, `encoder_p1_combiner`, `encoder_down_variant`, `decoder_gate_up_variant`, and a read-only `VLAModel.pipeline` accessor. Documented in `docs/stable_api.md`.

## Fixes made while porting

- **`S_kv_max > 1024` was silently wrong** in the seqused softmax fold: the register-tiled kernel caps at 1024 columns with no bound check, so wider logits rows kept an unnormalized tail that the PV GEMM still consumed, while the public binding accepted any length and its docstring advertised the limit. Rows past the tile now use a multi-pass kernel that holds no per-column registers, so any `S_kv_max` is correct; pinned at 1024 / 1025 / 2048 against torch SDPA.
- **`attention_seqused_fused.cu` and `rope_vec.cu` were compiled into the shared module on every architecture.** They join the existing SM100-class group (`FLASHRT_HAVE_THOR_VLA_KERNELS`), and their bindings are guarded by the same define so source and binding drop together on SM8x.
- **Local paths and internal references removed**: the harness carried a developer's checkpoint/fixture directories as literal defaults (now `--checkpoint` / `--fixture` or `$PI05_CHECKPOINT` / `$PI05_FIXTURE_DIR`, with a clear error when neither is set), and the doc's reproduction command and two "handoff reference in private issue" mentions are gone.
- **LingBot left alone**: the series had renamed every `sm_101a` mention to `sm_110a` across LingBot docs, example, comments and `pyproject.toml`. That is no longer accurate — the FA4 loader picks the chip string from the installed `nvidia-cutlass-dsl` — so those files are restored to main.

## Review follow-ups

- **The published configuration was not reachable from `load_model()`.** `use_fp4_encoder_attn` and `use_fp4_siglip_ffn` were not exposed at all, and `encoder_p1_combiner` defaulted to `lut_native` while the harness ran the fused GeGLU epilogue. Both are now public options and `use_fp4_decoder=True` resolves the measured preset. The harness gained `--construct load_model` (the new default), which builds both children through the public API and refuses to run when any sweep knob deviates from the preset; `--construct frontend` remains for the exploratory knobs and is labelled as not producing public-API numbers. `result.json` records the mode and the exact call. Re-measured in that mode, the action and raw cosines came out identical at every view count to the direct-construction runs the table above reports, which is the evidence the two build the same thing.
- **`use_fa4` did not match its documented contract.** The doc claimed it required `use_fp4=True`; the code allowed FP8 + FA4. The code was right — that combination is the measured baseline — so the doc is corrected and the independence is now stated in `docs/stable_api.md`, the `load_model` docstring and a comment at the check. Its config/framework/hardware validation also moved ahead of pipeline-class resolution, so an unsupported combination fails before a frontend import instead of after.
- **The NVFP4 GEMM return codes in the FP4 encoder FFN path were ignored**, including the fused GeGLU epilogue on the default path. They now raise a `RuntimeError` naming the kernel, layer and M/N/K. (The fused elementwise bindings in the same block validate in C++ and return nothing; those are deliberately not wrapped, and the helper's docstring says so.)
- **Minor**: the benchmark recorded `nvfp4_16_layers` while the SigLIP default covers all 27 — now recorded as the actual count; the `layer_norm_fp4_sfa_fp16` binding had no runtime caller and duplicated `layer_norm_mul_fp4_sfa_fp16` with a null inverse scale, so it is removed; the title now names the scope.

## Tests

`tests/test_pi05_fp4_fusion_kernels.py` and `tests/test_pi05_decoder_fp4_kernels.py` pin the FP4 kernels against the unfused chains they replace, including the new wide-logits boundary cases.

`tests/test_pi05_thor_fp4_routing.py` (new, 24 cases, no GPU needed) covers the `load_model` contract: FA4 reaching both frontends and being rejected elsewhere, each NVFP4 sub-flag requiring `use_fp4` and Pi0.5-torch-Thor, explicit values overriding the preset, and — the important one — the resolved constructor arguments asserted against the harness's own preset table, so the published configuration and the public API cannot drift apart. Two cases also assert that the knobs `load_model` does not forward still default to the published values in the frontend signature.

Full-suite regression on Thor: 747 passed, 198 skipped. The 25 failures in that run are pre-existing and unrelated (qwen36 GDN numerics, `test_groot_n17_e2e` needing the `gr00t` package, and test-ordering artifacts in `weight_loader` / `minimax` / `qwen3_vl` that pass in isolation); the identical set fails with this PR's new test file deselected.

## Reproduce

```bash
PI05_CHECKPOINT=<checkpoint-dir> PI05_FIXTURE_DIR=<fixture-dir> \
PYTHONPATH=<repo-root> python tests/bench_pi05_decoder_fp4_e2e.py \
  --num-views 3 --output-dir <output-dir>
```

The harness runs the FP8 and NVFP4 children in one session, writes both action tensors, and fails on any gate. It requires a clean tracked worktree so the recorded commit identifies the build.

## Base

Rebased onto current `main` (after #163). The Pi0.5 series and the GROOT N1.7 tier share the `csrc/gemm/fp4` and `csrc/fused_fp4` trees and the SM100-class kernel group; the rebase conflicts were all additive list entries plus the FA4 loader's chip-string selection, where main's version-aware implementation was kept.


LiangSu8899:
@heiheiha798 Nice tile push

PR #166, opened 2026-08-06
feat(chameleon): Chameleon-7B Thor FP8 and Orin SM87 INT8/QuaRot frontends

DXICM:
## Summary

Adds standalone Chameleon-7B image/text VLM frontends for Jetson Thor SM110 and Jetson Orin SM87.

#165 has been merged, and this branch is now rebased directly onto its merge commit on `main` (`20354060`). The PR contains only the Chameleon model, frontend, documentation, and test delta; it has no `csrc/` or CMake changes.

### Frontends

- Thor: dynamic FP8 backbone, optional FA4, eager Apache-2.0 Transformers VQ-VAE by default, explicit TensorRT VQ-VAE opt-in, and incremental KV-cache decode.
- Orin: INT8/QuaRot path with the FP16 causal FA2 support merged in #165.
- Registry entries stay lazy; `load_model(config="chameleon")` redirects callers to the direct chat-style frontend API.

### License boundary

The Research-License Meta VQGAN implementation, license, and notice are not included.

Both Thor eager tokenization and the TensorRT builder use the Apache-2.0 Transformers `ChameleonVQVAE` implementation and load only `model.vqmodel.*` tensors from the user-provided safetensors checkpoint. Loading is strict and shard-selective. The adapter supports both the Transformers 4.43 tuple output and newer `image_tokens` output.

The `chameleon` optional dependency group installs Torch, safetensors, Transformers, and Pillow.

### Runtime contracts

- Thor rejects `max_seq < 16` before allocation and validates padded prompt capacity.
- Thor and Orin perform architecture fail-fast checks before checkpoint loading or allocation.
- Negative generation lengths are rejected; zero-length Orin generation returns without prefill/decode.
- Chameleon-ON tests expose missing symbols, dependency failures, and unexpected import/link errors.
- Optional `flash_rt_fp4` handling catches only the exact missing module.

### Packaging integration

The rebase preserves the current packaging contract:

- no compiled `.so` files are shipped in the wheel;
- `flash_rt.tests` is excluded;
- recursive kernel/config package-data rules remain unchanged;
- the existing `motus` extra remains present;
- `chameleon` is added to the aggregate `all` extra.

### Validation

- Focused install, generic binding, registry, model-loading, Chameleon contract, VQ-VAE backend, and strict loader suite: **102 passed, 4 skipped**.
- Chameleon-focused suite with Python 3.13 / Transformers 4.55.4: **27 passed**.
[...]

PR #165, opened 2026-08-06
feat(csrc): Chameleon-7B kernel layer — fused quant, INT8 GEMMs, FA2 causal attention

DXICM:
## Summary

Adds the opt-in csrc kernel layer required by the Chameleon-7B model integration. Model frontends remain in the dependent #166.

### Kernel surface

- FP16/FP8 fused quantization and INT8 rowwise norms
- SM80 INT8/INT4 rowwise CUTLASS GEMMs and QuaRot/Hadamard helpers
- Fused Chameleon QK norm + RoPE
- Orin FA2 FP16 causal attention
- Thor SM110 FP16/FP8 CUTLASS causal FMHA shared libraries

### Build isolation

All model-specific sources and bindings remain behind `FLASHRT_ENABLE_CHAMELEON`.

The Chameleon option is now propagated consistently through the FA2 vendor object and Python adapter. With `FLASHRT_ENABLE_CHAMELEON=OFF`:

- FP16 causal FA2 specializations are not compiled.
- The wrapper has no real or stub FP16 causal entry.
- `flash_rt_fa2` does not export `fwd_fp16_causal`.
- The module has no defined or undefined `fvk_attention_fa2_fwd_fp16_causal` reference.

With the option enabled, the specialization and Python API are restored together. The unused public `GemmRunner.autotune_fp8_nn_bias` API and implementation were removed.

### Validation

- SM87, Chameleon OFF: built `flash_rt_kernels` and `flash_rt_fa2`; both modules import successfully; `fwd_fp16_causal` and its ELF references are absent.
- SM87, Chameleon ON: FA2 adapter compiles and links with the FP16 causal specialization.
- RTX 5090 / SM120 native Chameleon-ON FA2 reference tests: `7 passed, 3 skipped`, covering prefill vs SDPA, decode, causality, and build contracts.
- SM110 Chameleon-ON cross-build: `flash_rt_kernels`, `libfmha_fp16_causal.so`, and `libfmha_fp8_causal.so` compile and link. The Python module imports, and both FMHA libraries contain SM110a cubins with their expected exports.
- Static build-contract tests: `2 passed, 5 skipped` without build artifacts.
- `compileall` and `git diff --check`: pass.

SM110 kernels were cross-built and inspected on an RTX host; no claim is made here that the SM110 numerical kernels were executed on the RTX device.

## Dependency

#166 contains the model/frontend layer and must be merged after this PR.

PR #167, opened 2026-08-06
feat(hyvla): Hy-Embodied-0.5-VLA Thor SM110 — dynamic FP8 + fused megakernels

DXICM:
## Summary

Adds Hy-Embodied-0.5-VLA to FlashRT on Jetson Thor SM110. The Orin SM87 port is tracked separately in #168 and is stacked on this branch.

This branch is rebased onto current `main` (`51b3d0b4`, including #165 and #166).

### Architecture and isolation

- `models/hyvla/pipeline_thor.py`: Thor compute path
- `frontends/torch/hyvla_thor.py`: `HyVLATorchFrontendThor`
- `frontends/torch/_hyvla_thor_spec.py`: declarative weight spec
- `_PIPELINE_MAP`: `("hyvla", "torch", "thor")`
- HyVLA kernels are disabled by default.
- Thor-only sources and bindings require `FLASHRT_ENABLE_HYVLA=ON` and `GPU_ARCH=110`.

### Review hardening

- FP8 quantization handles odd element counts, including the final BF16 value.
- Fused GQA validates `nq == nkv * kv_rep` and uses overflow-safe offset-window checks.
- State and noise contracts are validated before ViT, graph lookup, capture, or GPU work.
- `FLASHRT_HYVLA_FORCE_ARCH` bypasses the architecture check only for exact value `1`.
- No-GPU tests skip only an explicitly missing unbuilt extension; unrelated imports, undefined symbols, and link failures remain visible.

### Performance receipt

Contributor-reported Thor SM110 result with a real image:

| Metric | Value |
|---|--:|
| E2E predict | **159.6 ms** (~5.8x vs reference eager ~930 ms) |
| Action cosine vs reference | **0.999706** |

### Latest validation

- Rebased cleanly onto `main` after Chameleon #165/#166.
- `GPU_ARCH=110`, `FLASHRT_ENABLE_HYVLA=ON`, Chameleon/Motus OFF: configure and `flash_rt_kernels` build passed.
- Extension import passed; fused attention, FP8 quant, and FFN symbols are present.
- Focused HyVLA + model-loading suite: **90 passed, 3 skipped**.
- Changed Python `compileall` and `git diff --check` passed.

The validation host is not SM110. The odd-tail numerical test and real graph-safe/E2E tests remain correctly hardware-gated; final merge still requires the Thor device receipt to be rerun or accepted from the contributor's existing result above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1813,6 +1813,29 @@ target_include_directories(flash_rt_kernels PRIVATE
   ${CUTLASS_DIR}/tools/util/include
 )
 
+# ── Hy-Embodied-0.5-VLA Thor SM110 kernels ──
+if(GPU_ARCH STREQUAL "110")
+  target_sources(flash_rt_kernels PRIVATE
+    csrc/kernels/hyvla_fused_thor.cu
+    csrc/kernels/hyvla_vit_fuse.cu
+    csrc/kernels/hyvla_quant_fp8_thor.cu
+    csrc/kernels/hyvla_ffn_fp8_thor.cu)
+  target_compile_definitions(flash_rt_kernels PRIVATE FLASHRT_HAVE_HYVLA_THOR=1)
+  message(STATUS "Hy-VLA Thor SM110 kernels: ENABLED")
+endif()
+
+# ── Hy-Embodied-0.5-VLA Orin SM87 kernels ──
+# The fused RoPE+QKNorm+KV-write megakernel is plain CUDA (no SM110-only
+# instructions) and removes ~11 launches per attention block; the FP8
+# Thor kernels stay SM110-only.
+if(GPU_ARCH STREQUAL "87")
+  target_sources(flash_rt_kernels PRIVATE
+    csrc/kernels/hyvla_fused_thor.cu
+    csrc/kernels/hyvla_vit_fuse.cu)
+  target_compile_definitions(flash_rt_kernels PRIVATE FLASHRT_HAVE_HYVLA_ORIN=1)
+  message(STATUS "Hy-VLA Orin SM87 fused attention-prep kernel: ENABLED")
+endif()
+
 target_link_libraries(flash_rt_kernels PRIVATE
   CUDA::cublas
   CUDA::cublasLt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1813,8 +1813,12 @@ target_include_directories(flash_rt_kernels PRIVATE
   ${CUTLASS_DIR}/tools/util/include
 )
 
-# ── Hy-Embodied-0.5-VLA Thor SM110 kernels ──
-if(GPU_ARCH STREQUAL "110")
+# ── Hy-Embodied-0.5-VLA model kernels (opt-in) ──
+# HyVLA sources compile only when FLASHRT_ENABLE_HYVLA=ON together with the
+# matching architecture gate. HyVLA OFF leaves the default SM87/SM110 build
+# and symbol surface unchanged.
+option(FLASHRT_ENABLE_HYVLA "Build HyVLA model kernels" OFF)
+if(FLASHRT_ENABLE_HYVLA AND GPU_ARCH STREQUAL "110")
   target_sources(flash_rt_kernels PRIVATE
     csrc/kernels/hyvla_fused_thor.cu
     csrc/kernels/hyvla_vit_fuse.cu
@@ -1824,11 +1828,10 @@ if(GPU_ARCH STREQUAL "110")
   message(STATUS "Hy-VLA Thor SM110 kernels: ENABLED")
 endif()
 
-# ── Hy-Embodied-0.5-VLA Orin SM87 kernels ──
 # The fused RoPE+QKNorm+KV-write megakernel is plain CUDA (no SM110-only
 # instructions) and removes ~11 launches per attention block; the FP8
 # Thor kernels stay SM110-only.
-if(GPU_ARCH STREQUAL "87")
+if(FLASHRT_ENABLE_HYVLA AND GPU_ARCH STREQUAL "87")
   target_sources(flash_rt_kernels PRIVATE
     csrc/kernels/hyvla_fused_thor.cu
     csrc/kernels/hyvla_vit_fuse.cu)
--- a/csrc/bindings.cpp
+++ b/csrc/bindings.cpp
@@ -8634,7 +8634,7 @@ graph-replay safe) to fill the SMs on long K. M in 1..16; N%8==0; K%64==0;
                 throw py::value_error(
                     "hyvla_rope_qknorm_kvwrite_bf16 supports hd==128 only, "
                     "got " + std::to_string(hd));
-            if (off < 0 || off + S > S_tot)
+            if (off < 0 || S > S_tot || off > S_tot - S)
                 throw py::value_error(
                     "hyvla_rope_qknorm_kvwrite_bf16: invalid offset window "
                     "off=" + std::to_string(off) + " S=" + std::to_string(S) +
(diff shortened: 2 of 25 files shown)


LiangSu8899:
Thank you for contributing the native Hy-Embodied-0.5-VLA Thor path. The model/frontend/pipeline split follows the repository's intended structure, the kernel names are model-scoped, and the matching CMake definitions and pybind guards are a good start. Codex performed a strict review against FlashRT's long-term maintenance rules. The following items should be addressed before merge.

### Required changes

1. **Add a model-level build option, not only an architecture gate**

   The new HyVLA sources are currently compiled into the common `flash_rt_kernels` module for every SM110 build, and two of them are also compiled for every SM87 build. Architecture gating prevents unsupported code generation, but it does not provide model-selected compilation: users building unrelated models on Thor or Orin still pay the compile, link, binary-size, and symbol-surface cost.

   Please add an option disabled by default, for example:

   ```cmake
   option(FLASHRT_ENABLE_HYVLA "Build HyVLA model kernels" OFF)
   ```

   Gate the sources, compile definitions, and pybind symbols together. The expected matrix is HyVLA OFF on all architectures, HyVLA ON + SM110 for the Thor kernel set, and HyVLA ON + SM87 for only the shared portable kernels.

2. **Fix the public `load_model(..., use_fp4=True)` route**

   In the HyVLA FP4 branch, `pipe_cls` is changed to `HyVLATorchFrontendThor`, but `use_fp4` is then set to `False`. The generic kwarg construction does not pass `use_fp4=True` to this frontend, so the advertised HyVLA FP4 route actually constructs the default non-FP4 frontend.

   Please preserve a dedicated HyVLA FP4 routing flag and pass it explicitly to the selected frontend. Add a routing test that inspects or mocks the constructed frontend and proves that `use_fp4=True` reaches it. The standard `load_model` path should also have an explicit, tested way to select the validated `use_fused` production tier if that is part of the supported API.

3. **Fail early on unsupported hardware and avoid implicit remote-code execution**

   `HyVLATorchFrontendThor` does not validate SM110 before loading weights and allocating CUDA state. Please check `torch.cuda.get_device_capability()` at the beginning of initialization and report a clear error on unsupported devices.

   The tokenizer currently uses `trust_remote_code=True`. That allows checkpoint-provided Python code to execute during model loading and conflicts with the stated self-contained runtime boundary. Please use the standard tokenizer path without remote code. If custom remote code is genuinely required, make it an explicit, documented opt-in rather than the default.

4. **Validate the public fused-kernel contracts before launch**

   The pybind APIs accept raw pointers and unconstrained dimensions, while the implementations rely on production-shape assumptions. At minimum, validate the supported HyVLA contract before launch:

   - fused RoPE/QK-Norm/KV-write: positive `S`, `nq`, `nkv`, `S_tot`; supported `hd` (currently 128); valid `off` and `off + S <= S_tot`; positive `eps` and `kv_rep`
   - ViT add + LayerNorm: positive rows, supported/even dimension, positive `eps`
   - FFN kernels: positive dimensions, required `K` alignment, and `N`/`Nout` divisibility by the 32-column tile

   Unsupported inputs must raise a Python exception rather than launch invalid grids or silently leave output columns unwritten.

5. **Commit the validation that the documentation claims**

   `docs/hyvla05_thor_sm110.md` references multiple precision, graph-safety, and stage tests under `tests/test_thor_hyvla05_*`, but none of those files are present in this PR. The PR itself adds no test file, and every test-plan item is currently unchecked. Performance and cosine numbers are useful evidence, but they need a reproducible in-tree gate.

6. **Tighten the frontend input contract**

   `predict_actions()` uses an `assert` for the missing-prompt state and does not reject state vectors larger than `max_state_dim` or invalid noise shapes. Please replace the assertion with a stable exception and validate state/noise/camera shapes before graph lookup or copy. Public API errors should be deterministic in both normal and `python -O` execution.

### Pre-merge checklist

- [ ] `FLASHRT_ENABLE_HYVLA=OFF` leaves the default SM87/SM110 build and symbol surface unchanged
- [ ] HyVLA ON + SM110 build/import/symbol checks pass
- [ ] HyVLA ON + SM87 exposes only the intended shared portable symbols
- [ ] `load_model(..., config="hyvla", use_fp4=True)` demonstrably enables the FP4 frontend path
- [ ] Thor rejects non-SM110 devices before checkpoint loading
- [ ] Tokenizer loading does not execute remote checkpoint code by default
- [ ] Fused kernels have host-side contract validation and numerical reference tests
- [ ] Prompt, state, noise, image-shape, eager/graph, and replay-stability tests are committed
- [ ] Claimed fixed-noise action cosine and graph-safety gates are reproducible from files in this repository
- [ ] `git diff --check`, Python compilation, default build, and HyVLA smoke tests pass

Please refer to the repository's [PR Review Checklist](https://github.com/flashrt-project/FlashRT/blob/main/docs/pr_review_checklist.md) and [Adding a New Model](https://github.com/flashrt-project/FlashRT/blob/main/docs/adding_new_model.md) guides.

This is a Codex-assisted maintainability review. The architecture and optimization direction are reasonable; the model-selected build boundary, public routing correctness, safe input contracts, and reproducible validation need to be closed before this becomes a supported path on `main`.

DXICM:
Thanks for the review. All six required changes are addressed, and re-validated on a production Jetson Orin (sm_87, CUDA 12.2, torch 2.3.0).

**1. Model-level build option** — `option(FLASHRT_ENABLE_HYVLA "Build HyVLA model kernels" OFF)` added; sources, compile definitions, and pybind symbols are gated together. Matrix verified on SM87:
- OFF: default build/symbol surface unchanged (no HyVLA symbols).
- ON + SM87: only the shared portable kernels export (`hyvla_rope_qknorm_kvwrite_bf16`, `hyvla_vit_add_layer_norm_bf16` = True; Thor-only `hyvla_quant_fp8_dyn_bf16` / `hyvla_ffn_gu_silu_bf16` = False).
- ON + SM110: full Thor kernel set (build-gated; SM110 device run queued for the Thor hardware).

**2. `load_model(..., use_fp4=True)` route** — a dedicated HyVLA FP4 routing flag now carries through `flash_rt/api.py` and explicitly passes `use_fp4=True` to the selected frontend. `tests/test_hyvla_fp4_routing.py` mocks the constructed frontend and proves:
- `use_fp4=True` reaches the frontend;
- the validated FP8 production tier selects `use_fused=True`;
- the default route does not enable FP4;
- SM87 FP4 degrades to the INT8 path with an explicit warning instead of silently claiming FP4.
Production: 3 passed, 1 skipped.

**3. Fail-fast + no implicit remote code** — `HyVLATorchFrontendThor` validates `torch.cuda.get_device_capability()` at the top of initialization (before checkpoint loading and CUDA allocation) and raises a clear `RuntimeError` on non-SM110 devices; dev override is the documented, HyVLA-specific `FLASHRT_HYVLA_FORCE_ARCH` env var. The tokenizer loads without `trust_remote_code`. Covered by `tests/test_hyvla_arch_gate.py` (4 passed).

**4. Fused-kernel contracts** — host-side validation added before launch for all five fused kernels (RoPE/QK-Norm/KV-write, ViT add+LayerNorm, dynamic FP8 quant, FFN gu/silu, FFN dn/res): positive dims, supported `hd == 128`, valid `off + S <= S_tot`, positive `eps`/`kv_rep`, required K alignment, and 32-column tile divisibility. Invalid inputs raise Python exceptions instead of launching invalid grids. Covered by `tests/test_hyvla_kernel_contracts.py` (5 passed, 3 skipped on SM87 where the Thor-only symbols are correctly absent).

**5. Committed validation** — the PR now ships `test_hyvla_thor_dispatch.py` (3 passed), `test_hyvla_arch_gate.py` (4 passed), `test_hyvla_fp4_routing.py` (3 passed / 1 skipped), `test_hyvla_kernel_contracts.py` (5 passed / 3 skipped), and `test_hyvla_thor_graphsafe.py`. `docs/hyvla05_thor_sm110.md` was updated to reference only committed files.

**6. Frontend input contract** — the missing-prompt `assert` is now a stable `RuntimeError` (deterministic under `python -O`); state/noise shapes are validated before graph lookup and copy, with oversized state vectors rejected.

Combined production run: **43 passed / 0 failed / 5 skipped** (all skips are intended hardware gates). Remaining SM110-device checklist items (Thor build/import/symbol checks, fixed-noise cosine and graph-safety gates on real Thor hardware) are queued for the Thor device and will be reported here.

LiangSu8899 (approved):

PR #169, opened 2026-08-06
Qwen3.6-35B-A3B edge runtime: white-boxed prefill, SM110 support, and token-exact speculative decode

LiangSu8899:
Brings the Qwen3.6-35B-A3B (`qwen3_5_moe`) text path from a working SM120
runtime to one that also runs on Jetson AGX Thor, white-boxes the prefill onto
its own kernels, and adds speculative decode that is token-exact rather than
approximately so.

The branch is one model path end to end. It is large; the sections below are
ordered so a reviewer can stop after any of them.

## What changes

**Prefill, white-boxed.** The prefill was reaching for tensor ops at 202 call
sites. Those are now kernels: fused GDN WY pack/norm/cumsum, a row-blocked
causal convolution that takes its own history so the chunked path stops
concatenating, a five-kernel MoE routing producer, and a fused shared-gate
combine. What the profiler still attributes to tensor ops at a 2048-token
prompt is 0.7 ms of KV-cache writes, against 29% of the prefill before.

**FA2 on SM110.** The designated Thor attention path materialises an
`(S*heads, S_kv)` score buffer -- gigabytes per layer at long context. FA2 was
excluded from the arch list on build size, not capability, and the causal
hd256 instantiation was already in the tree. Enabling it makes the non-square
window of a chunked prefill 20x faster and takes the chunking penalty from 64%
to 4%. It is opt-in (`-DFLASHRT_ENABLE_THOR_FA2=ON`), so no other Thor build
pays for it.

**Decode kernels.** Five changes, each bit-identical to what it replaced:
gating constants derived once instead of every step; the MoE tail fused into
one kernel; the gated-DeltaNet recurrence stopped spilling its state column to
local memory; the router's top-8 picked in one warp instead of `k` rounds of
block-wide barriers; and the W4A16 GEMV's loads-in-flight halved, which trades
per-thread depth for occupancy the profiler said was register-limited. That
last one is **scoped to sm_110 in CMake**: it was measured on a 20-SM part and
a larger device may prefer the deeper parallelism, so every other target
compiles the constant it compiled before.

**Speculative decode.** A DeepSeek-V3-style MTP head, verified through the
decode kernels at `K+1` rows over the weights the decode step caches, so a
verified row *is* the decode step it stands in for. Enabled with
`load_mtp=True`; `generate()` never reads the head.

**A prefill reduction that was not deterministic.** Three of the four MoE paths
summed each token's experts with `index_add_`, which reduces through atomics.
The same prompt did not decode to the same text twice, and the 16-token fixture
never reached the divergence. The fourth path already had a fixed-order kernel
and said why in a comment; the other three now match it.

### Jetson AGX Thor (sm_110) vs vLLM 0.26.0

Same part, same BF16 checkpoint, same pre-tokenized prompts, same protocol on
both sides, one length per process. Vision tower off on both.

| prompt | TTFT | vLLM TTFT | | decode | vLLM decode | |
|---:|---:|---:|---|---:|---:|---|
| 20 | **89.5 ms** | 102.3 ms | +14% | | | |
| 256 | **104.7 ms** | 214.5 ms | +105% | | | |
| 512 | **144.5 ms** | 251.1 ms | +74% | | | |
| 1024 | **216.0 ms** | 319.4 ms | +48% | **87.1 tok/s** | 31.6 tok/s | 2.8x |
| 2048 | **379.6 ms** | 495.0 ms | +30% | **86.3 tok/s** | 31.5 tok/s | 2.7x |
| 4096 | **748.6 ms** | 867.4 ms | +16% | **85.1 tok/s** | 31.2 tok/s | 2.7x |
| 10240 | **1890.7 ms** | 2144.5 ms | +13% | | | |
| 32768 | **7207.5 ms** | 7231.8 ms | +0.3% | | | |

TTFT is ahead at every length measured. 128 K context reaches this board at
2470 tok/s of prefill, which it could not before. The decode column was taken
before the round below, which added 15.3% to this side and nothing to vLLM's,
so those ratios are lower bounds.

### What moved on Thor

| | before | after | |
|---|---:|---:|---|
| TTFT @2048 | 532.0 ms | **379.6 ms** | -29% |
| decode step | 11.238 ms | **9.743 ms** | +15.3% (89.0 -> 102.6 tok/s) |
| prefill reached through tensor ops @2048 | 29% | **0.7 ms** | |

Every decode change is bit-identical to what it replaced.

### Speculative decode (Thor)

Plain and speculative in the same process, emitting plain greedy's sequence
token for token with the 16-token golden fixture passing on the same build.

| | tok/s | vs plain |
|---|---:|---:|
| plain greedy | 100.35 | |
| speculative, K=1 | 105.22 | 1.05x |
| speculative, K=2 | **106.74** | **1.06x** |

For scale: vLLM runs the same MTP head and gets a larger relative gain --
31.59 -> 55.00, 1.74x -- because its step is three times heavier, so a fixed
per-draft cost is proportionally three times cheaper. Plain greedy decoding
here still beats that speculative figure by 1.8x, with no speculation at all.

### SM120

**No new SM120 latency claim is made.** The card available for it is shared
with other work. The reference stays the same-shape `P=64, N=64` warm-graph
measurement already on record for this path:

| Measurement | Result |
|---|---:|
| Prefill | 40.42 ms |
| Decode, median of 8 runs | 257.95 tok/s |
| Decode, range across 8 runs | 256.90-258.22 tok/s |
| Repeated sequences identical | 8 / 8 |

The tuned constant discussed above is scoped to sm_110, so the SM120 kernels
are byte-identical to the ones that measurement covered. Per-change breakdowns
for both architectures are in `docs/qwen36_moe_usage.md`.

## Correctness

Every kernel change is checked with `torch.equal` against the path it replaces,
not a tolerance, because the emitted tokens are compared against a fixture and
the speculative window is only allowed to keep a token because it computed what
the decode step would have.

| what | how | result |
|---|---|---|
| speculative window vs decode steps | logits, per-token recurrent and conv snapshots, written KV rows, over three windows of four at real decoded tokens | bit-identical |
| gated-DeltaNet recurrence | six trials of eight *chained* steps -- a recurrence that drifts by an ulp does not show in one step | outputs and final state bit-identical |
| router top-8 | 800 inputs across four widths, 397 with a tie inside the top-8 | indices and values identical |
| fused shared-gate combine | every shape and scale the two paths issue | identical |

The combine was *not* identical at first: the compiler contracts the multiply
and add into one fma, one rounding where the tensor-op chain has two, and that
showed up as one element in 16384 differing by one ulp. Fixed with explicit
rounded operations, which also removed a prefill-vs-decode disagreement that
had gone unnoticed.

## Build boundaries

A build with every `qwen3_5_moe` option off compiles the same sources and
exports the same symbols it did before this branch. Three additions did not
respect that and were reworked:

| addition | was | now |
|---|---|---|
| grouped NVFP4 MoE GEMM | a second source in the SM100 W4A16 object every Thor build compiles | its own object library, `GPU_ARCH=110 AND FLASHRT_ENABLE_QWEN35MOE_W4A16`, bindings on `FLASHRT_HAVE_QWEN35MOE_GROUPED_SM100` |
| grouped activation quantisers | appended to the shared `quantize.cu` under the broad NVFP4 gates | `csrc/kernels/qwen35moe_grouped_quant.cu`, built with the weight-only tier |
| FA2 on sm_110 | unconditional for the architecture | `FLASHRT_ENABLE_THOR_FA2`, default OFF |

The element and scale converters the grouped quantisers share with the general
quantiser moved to a header rather than being copied, so `quantize.cu` compiles
the same definitions it did.

`scripts/qwen35moe_build_matrix.py` prints the sources and symbols each tier
adds and fails if any of them is reachable with every tier off;
`tests/test_qwen35moe_build_matrix.py` runs the same checks. The five
configurations behind it, all verified configure-only:

| configuration | result |
|---|---|
| `-DGPU_ARCH=120` | `FA2 ENABLED`; no `qwen3_5_moe` source or symbol |
| `-DGPU_ARCH=110` | `FA2 DISABLED`; no `qwen3_5_moe` source or symbol |
| `-DGPU_ARCH=110 -DFLASHRT_ENABLE_QWEN35MOE_CORE=ON -DFLASHRT_ENABLE_QWEN35MOE_W4A16=ON -DFLASHRT_ENABLE_THOR_FA2=ON` | core + weight-only tiers, grouped MoE GEMM, FA2 at `hdim={256} x dtype={bf16}` |
| `-DGPU_ARCH=120 -DFLASHRT_ENABLE_QWEN35MOE=ON` | all three tiers |
| `-DGPU_ARCH=110 -DFLASHRT_ENABLE_QWEN35MOE_W4A4=ON` | **configure fails**, naming the two tiers that do apply |

## Runtime boundaries

- **Thor is a registered path, not a partial one.** The frontend moved to
  `flash_rt.frontends.torch.qwen36_moe` as `Qwen36MoeTextFrontend` -- nothing in
  it is architecture-specific -- and both `("qwen36_moe", "torch", "thor")` and
  `("qwen36_moe", "torch", "rtx_sm120")` resolve to that one class, with a
  resolver test asserting they are the same object. The previous module and
  class name re-export it.
- **Speculative decode has a supported API.** `load_mtp=False` by default on
  the constructor, validated, documented with a complete example, and covered
  by tests. It was previously reachable only by setting a private attribute on
  a subclass, which the error message then recommended.
- **Its graph cache no longer borrows the decode cap of 256.** A speculative
  window covers `k+1` positions through the whole stack, so its memory pool is
  several times a decode step's; 256 of them is more than a 32 GB board has at
  a 2048-token context, which is where it was measured running out. The default
  is 16, `spec_graph_cache_max` sets it, and the eviction policy is tested.
- **Importing the shared forward no longer mutates the environment.** It set
  `FLASHRT_BF16_CUBLASLT_AUTOTUNE_ALGOS`, which is process-global and shared
  with every frontend, so a model loaded later inherited an autotune setting it
  never asked for. The GEMM entry takes the bound as an argument instead
  (default 0 = the existing behaviour), and plans are cached per
  `(M, N, K, max_algos)`.
- **Kernel selection is a policy, not a symbol lookup.** `KernelPolicy` gathers
  the choices this path makes -- edge W4A16, the routing producer, the WY scan,
  the fused combine, the warp router, the edge recurrence, the verify form --
  with the existing environment variables as its defaults. Every field selects
  between implementations checked against each other with `torch.equal`, so a
  field decides speed and cannot decide output. Nex-N2 regression tests assert
  the shipped defaults field by field and that turning a field off selects the
  fallback even when the faster symbol is present.
- The build errors name the tiers the device in front of the reader can
  compile, rather than recommending the switch that turns on the block-scaled
  tier -- which sm_110 refuses at configure time.

## Validation

SM110, Jetson AGX Thor, built with the supported tier combination
(`-DGPU_ARCH=110 -DFLASHRT_ENABLE_QWEN35MOE_CORE=ON
-DFLASHRT_ENABLE_QWEN35MOE_W4A16=ON -DFLASHRT_ENABLE_THOR_FA2=ON`), against the
official BF16 checkpoint:

```
kernel preflight (M-row W4A16, fused combine, conv1d rows,
                  WY prefill, routing producer)          68 passed
speculative decode contract + equivalence                25 passed
build matrix / kernel policy / registry / contract       48 passed
gated-DeltaNet recurrence equivalence + timing           PASS, 1.68x
router top-8 equivalence + timing                        PASS, 1.34x
```

The speculative file is the one that matters for this branch's headline
feature: K=1 and K=2 against plain greedy, the window's emitted argmax and its
recurrent, conv and KV state against the decode steps it stands in for, the
rejected-tail rewind, boundary token counts, and the graph cache staying within
its bound -- all against the checkpoint, all equality rather than tolerance.

SM120, RTX 5090, PyTorch 2.9.1 / CUDA 12.8, BF16 checkpoint:

```
cmake -B build -S . -DGPU_ARCH=120 -DFLASHRT_ENABLE_QWEN35MOE=ON
cmake --build build --target flash_rt_kernels
FLASHRT_QWEN36_MOE_CKPT_DIR=/path/to/Qwen3.6-35B-A3B PYTHONPATH=. \
  PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -q tests/test_qwen36_moe_gpu.py
    -> 2 passed  (official Transformers BF16 greedy-token fixture)
```

The remaining SM120-visible changes -- the fused MoE tail, the router top-8,
the recurrence rewrite -- are each checked bit-for-bit against the path they
replace, and the golden-fixture gate above is the end-to-end confirmation.

CPU-side, no GPU required:

```
tests/test_qwen36_moe_smoke.py         registry, both architectures, checkpoint contract
tests/test_qwen36_moe_spec_decode.py   constructor contract, graph-cache eviction
tests/test_qwen35moe_build_matrix.py   tier gates, no leak with every tier off
tests/test_nexn2_kernel_policy.py      shared defaults, no import side effect
```

`benchmarks/qwen36_moe_edge_decode.py` gives the doc's table a reproduction
command; it reports a median and range over every repetition and refuses to
print throughput unless the eager and captured paths emit identical tokens.

## Notes for review

- Both new decode kernels are optional. The frontend resolves each through the
  policy above and falls back to the kernel it replaces, so a build without
  them is slower and never different. Neither is in the required-symbol set;
  the model doc lists both with the symbol they fall back to.
- Every new binding sits inside the same compile-time guard as the kernel it
  extends, and the implementing translation units are compiled under exactly
  those gates.
- The new gated-DeltaNet entry is shape-specialized to a head dim of 128 and
  raises with the operation name and shapes for anything else, rather than
  returning quietly and leaving the output undefined.
- `exec/` is untouched.
- Two doc claims were stale and are corrected: speculative decode is
  implemented and reachable, and FA2 on Thor is opt-in rather than absent. The
  root README's two RTX-only FA2 statements are updated to match.

## Known gaps

- Speculative acceptance rises with context but the ratio does not, for the
  reason given above. Closing it means a batched verify attention, which is a
  different kernel from the one the decode step uses -- and using the decode
  step's kernel is what makes the window bit-exact.
- At 32768 tokens the TTFT lead is +0.3%, i.e. level. That length is dominated
  by attention's O(S^2), which is the one component still reached through
  torch; the sweep's shape says so directly.
- The largest remaining prefill kernel runs at 15% occupancy and 1.6 waves per
  SM. The fix is a real retiling, not a constant, and is not in this PR.
- The decode column of the vLLM comparison predates the decode round, so its
  2.7-2.8x is a lower bound rather than a measurement of the current tree. Only
  this side needs re-sweeping; vLLM's is a different binary.


LiangSu8899:
Thank you for the substantial work here. The white-boxed prefill path, tiered Qwen3.5/3.6 build options, defensive capability probing, deterministic-state work, and the amount of real-device validation are all valuable contributions. I also checked the CPU-side test surface locally: `git diff --check` and Python compilation pass, and the Qwen3.6 smoke/quant/cache tests pass (`54 passed, 1 skipped`).

This is a **Codex-assisted maintainability review** against FlashRT's long-term contribution rules. Because this PR is large and touches shared build/runtime surfaces, I do not think it is ready to merge yet. The remaining issues are mostly boundary and integration issues rather than objections to the feature itself.

## 1. Keep the default build unchanged when Qwen3.5/3.6 is disabled

The new `_CORE`, `_W4A16`, and `_W4A4` tiers are a good direction and default to `OFF`. However, several new costs bypass those model gates:

- SM110 now enables the vendored FA2 target unconditionally. This changes compile time, binary size, and supported symbols for every Thor build, including unrelated models.
- `cutlass_nvfp4_moe_grouped_sm100.cu` is added to the existing SM100 W4A16 object and its bindings use the broad `ENABLE_CUTLASS_SM100_NVFP4_W4A16` guard.
- Qwen/MoE grouped NVFP4 quantization is added to the shared `csrc/kernels/quantize.cu`, with symbols exported under broad NVFP4/SM100 guards.

Please move the grouped MoE GEMM and grouped quantization implementation into dedicated Qwen3.5/3.6 translation units, and gate their sources, declarations, and bindings with the corresponding model tier. SM110 FA2 should likewise require a Qwen model-specific option (or a dedicated default-OFF Thor-FA2 option). A build with every Qwen3.5/3.6 option disabled should retain the previous source list and public symbol surface.

## 2. Complete the Thor integration instead of exposing it as a partial path

The PR documents Thor/SM110 as supported, but the public registry still only registers `qwen36_moe` for `rtx_sm120`; there is no `("qwen36_moe", "torch", "thor")` entry. The surrounding comments also still describe this family as RTX 5090 only.

Please add an explicit Thor frontend/registration and a resolver test. If the implementation is genuinely shared, a hardware-neutral base is preferable to requiring users to directly construct an `Rtx`-named frontend for Thor.

The build UX also needs to be exact:

- `FLASHRT_ENABLE_QWEN35MOE=ON` enables W4A4 and therefore intentionally fails on SM110.
- The Thor documentation currently contains `-DGPU_ARCH=110 ...` rather than a usable command.
- Missing-kernel errors still tell all users to rebuild on SM120 with the total switch.

Please document the exact SM110 tier combination and make errors architecture-aware. Do not recommend the total switch on a target where it is known to fail.

## 3. Provide a supported speculative-decode API and a safe cache policy

`generate_spec()` is public, but the frontend constructor has no `load_mtp` argument. The only documented way to enable it is to set the private `_load_mtp = True` attribute on a subclass, and the runtime error recommends the same private mechanism. This leaves the headline feature unavailable through the supported constructor contract.

Please add a public, default-OFF option such as `load_mtp: bool = False`, validate it, document a complete usage example, and cover it with a public API test.

The speculative graph cache also reuses `GRAPH_CACHE_MAX=256`, while the PR notes that this policy is too large for the named 32 GB target at a 2048-token context. Please give speculative graphs a separate conservative or memory-aware cap and test eviction. A supported optional feature should not default to a known OOM configuration on its primary target.

## 4. Isolate Qwen3.6/Thor policy from existing Nex-N2 behavior

This PR makes large changes in shared Nex-N2 modules (`_nexn2_rtx_forward.py`, `_nexn2_rtx_decode.py`, and `attn_backend_nexn2.py`). Several optimizations become default based on symbol presence or environment defaults, so existing Nex-N2 behavior can change even when users did not select the new Qwen3.6 path.

Most importantly, importing `_nexn2_rtx_forward.py` now executes:

```python
os.environ.setdefault("FLASHRT_BF16_CUBLASLT_AUTOTUNE_ALGOS", "1")
```

That mutates process-global configuration merely by importing a shared frontend module and can affect models loaded later in the same process. Please pass this as frontend/runner configuration rather than changing `os.environ` at module import.

Please also scope Qwen3.6/Thor performance policy (edge W4A16, router/GDN selection, fused combine, and related defaults) through an explicit frontend/config object. Common deterministic correctness fixes can remain shared when justified, but they need Nex-N2 regression coverage. Existing model defaults should not change because a newly built symbol happens to be present.

## 5. Reconcile the documentation with the implementation

There are direct contradictions that need one authoritative policy:

- `docs/qwen36_moe_usage.md` says this interface does not execute MTP and speculative decode is not part of it, then later documents `generate_spec()`.
- The same document says FA2 is deliberately absent on Thor, while this PR enables and tests FA2 for SM110.
- The root README still says `flash_rt_fa2.so` is RTX-only and Thor skips it in multiple places.

Please update the root README, model guide, frontend docstrings, build table, and error messages together after deciding the final gated FA2 policy.

## 6. Commit reproducible integration evidence

The added kernel tests are useful, but I could not find a committed test for the new speculative runtime contract. Please add a reproducible test/harness covering:

- construction through the public MTP option;
- K=1 and K=2 output equivalence with plain greedy decode;
- logits, recurrent/conv state, and KV-cache equivalence;
- rejected-tail rollback;
- zero/negative/capacity-boundary token counts;
- speculative graph-cache eviction.

Checkpoint/GPU tests may be skipped when prerequisites are absent, but the harness and assertions should live in the repository. Please also attach a build/symbol matrix for:

- SM110 and SM120 with all Qwen tiers disabled (unchanged baseline);
- SM110 with the supported CORE/W4A16/Thor attention combination;
- SM120 with all supported tiers;
- expected SM110 W4A4 configure rejection;
- symbol presence/absence for each tier.

### Merge checklist

- [ ] No Qwen-specific source or binding enters a default build when all Qwen3.5/3.6 tiers are `OFF`.
- [ ] Thor has an explicit registry path, exact build command, and resolver test.
- [ ] Speculative decode is enabled through a public API with a bounded default memory policy.
- [ ] Shared Nex-N2 defaults remain stable; there is no process-global import side effect.
- [ ] README/model docs/error messages agree on MTP, FA2, architectures, and build tiers.
- [ ] Speculative equivalence/rollback/cache tests and the SM110/SM120 build matrix are reproducible.

Once these boundaries are closed, the feature can remain substantial without imposing compile, symbol, runtime-policy, or memory costs on unrelated models and platforms.

LiangSu8899:
Rebase and Qwen3.6-35B-A3B SM110 validation update

The branch has been squash-rebased onto current `main` (`8dc4732`) and the shared conflicts were resolved manually. GitHub reports `MERGEABLE / CLEAN`.

The Qwen3.6-35B runtime remains split into `CORE`, `W4A16`, and `W4A4` tiers. The dense W16A16 implementation is compiled once behind a shared capability gate, avoiding duplicate translation units and bindings.

During the real Thor build, I found and fixed one additional issue: the slim Thor FA2 matrix builds BF16 hdim-256 only, but the Python causal wrapper still referenced the hdim-128 specialization unconditionally. That allowed linking but caused `flash_rt_fa2` import to fail with an undefined symbol. The wrapper now uses `FA2_HAS_HDIM_*`-guarded dispatch for both native and Python adapters, with a regression contract test.

Validation was performed in a clean, isolated Jetson AGX Thor (SM110) workspace using PyTorch 2.11.0+cu130, CUDA 13.0, and the official **Qwen3.6-35B-A3B BF16 checkpoint**.

Build and kernel validation:

- Full SM110 configure and build passed for `flash_rt_kernels` and `flash_rt_fa2` with Qwen3.6 `CORE=ON`, `W4A16=ON`, `W4A4=OFF`, and Thor FA2 enabled.
- Both native modules import successfully.
- Expected Qwen3.6 CORE/W4A16/grouped-SM100 symbols are present; all four W4A4 symbols are absent.
- Qwen3.6 kernel, build-matrix, registry, checkpoint-contract, and hdim-256 causal FA2 numerical tests passed, including square and non-square bottom-right causal windows.
- Official-checkpoint speculative suite: `26 passed`, covering MTP opt-in loading, K=1/K=2 equality with plain greedy, bit-exact recurrent/conv/KV state, rejected-tail rollback, boundary token counts, and graph-cache bounds.
- Changed-Python compilation and `git diff --check` passed.

End-to-end Qwen3.6-35B validation:

- Input boundary: a real user chat message.
- Timed boundary: chat-template formatting/tokenization -> prompt setup -> prefill -> all 16 decode steps -> CUDA synchronization -> final tokenizer decode.
- One-time model loading/runtime quantization was measured separately at `56.69 s` and excluded from warm request latency.
- Logits shape: `[1, 20, 248320]`.
- All logits finite: `true`.
- First argmax: `8160`, matching the official BF16 golden first token.
- Greedy generation: `5/5` runs exactly matched the official 16-token BF16 golden sequence.
- Greedy repeat determinism: `5/5` identical.
- MTP speculative K=2: `5/5` runs exactly matched plain greedy.
- Speculative repeat determinism: `5/5` identical.
- Warm greedy E2E P50: `218.71 ms` (range `218.63-219.64 ms`).
- Warm speculative K=2 E2E P50: `203.27 ms` (range `203.02-203.97 ms`).
- Final decoded text was finite/coherent and began: `Here's a thinking process:`.

The model and numerical conclusions in this report are exclusively for **Qwen3.6-35B-A3B**. No other model is presented as part of this validation, and no private paths are included.
Final attention-contract follow-up (`20d2306b`):

- Optional FA2 loading now suppresses only `ModuleNotFoundError` for `flash_rt.flash_rt_fa2` itself. Undefined symbols, missing transitive dependencies, and other import/link errors propagate.
- GPU tests no longer convert arbitrary frontend or extension import failures into skips.
- Automatic FA2 probe fallback emits a `RuntimeWarning`; an explicit `use_fa2=True` or explicit decode-FA2 environment request fails fast if the module is absent or the probe fails.
- The SDPA fallback now uses an explicit bottom-right causal mask for `1 < q_seq < kv_seq`; `q_seq > kv_seq` is rejected.
- A non-square fallback case (`Sq=3`, `Sk=7`) is checked numerically against an explicit fp32 bottom-right reference.

Post-fix Thor validation, exclusively on Qwen3.6-35B-A3B paths:

- Attention/import contract and real Thor FA2 numerical suite: `13 passed`.
- Official Qwen3.6-35B-A3B BF16 checkpoint speculative suite: `25 passed`.
- Full request E2E after the fix: logits `[1, 20, 248320]`, all finite, first argmax `8160` matching the official golden.
- Greedy: `3/3` runs exactly matched the official 16-token BF16 golden sequence.
- MTP K=2: `3/3` runs exactly matched greedy.
- Warm E2E P50: greedy `219.12 ms`; speculative K=2 `203.27 ms`.
- Final decoded text remained coherent.

PR #174, opened 2026-08-14
structures: LTX-2.5 attention forms and host bindings

LiangSu8899:
Brings the LTX-2.5 transformer to the structures layer as an ordinary Diffusers host: two executable forms of `attention_core`, the host-family adapter that recognises its attention seam, and the region and pipeline bindings. The model-private runtime is #172; nothing here depends on it.

## What this adds

- **`adapters/diffusers_gated_rotary_attention.py`** — the attention seam of hosts whose processors take separate query/key rotary boundaries, RMS-normalise Q/K after projection, and gate the output with per-head sigmoid logits. The stock Diffusers adapter's processor-state contract does not exist on these modules, so it correctly declines them. Recognition is structural: slots, the processor signature, and the declared rope form. No model or class name is discovery evidence.
- **`impls/attention_core/sage2_blackwell.py`, `sage3_blackwell.py`** — the INT8-QK and FP4 forms, reading their envelope from the installed kernel package. Both qualify through the family's capture convention and answer its way: `None` for an unclaimed shape, an exception for an inconsistent calibration or an unserved device.
- **`bindings/ltx25_dit.yaml`, `ltx25_video_pipeline.yaml`** — the region binding for the video feed-forward and complete hot-path coverage for the joint audio+video pipeline: 17 segments, four owned by catalog structures and the rest classified as host stages, state regions or control, with the distilled single-pass recipe, encoder residency and decode tiling budget as binding attributes.

## Selection is explicit, and the default order does not move

The quantized forms trade a bounded numerical error for speed, which is a decision about a deployment rather than about a device, so they do not enter the published precision-first order that existing hosts hold receipts against. They are, however, meant to be easy to try: the gate measures accuracy and latency and prints both, and `detach` reverses the attachment exactly, so trying one is how a deployment finds out rather than something it commits to.

They arrive on the axis that already carries such decisions: `QuantScheme` gains `attention_forms`, empty on every existing profile, read by attention adapters that declare `scheme_aware` exactly as the gated-delta adapter reads its projection format. `"nvfp4_balance_sage"` registers the profile that names them. `attach` and `auto_swaps` take `attention_forms` directly, which outranks the profile, so tuning one seam does not require authoring a scheme; `bind_dense_attention_best` gains the matching `prefer` argument for a caller assembling an adapter by hand.

The staging is pooled per (shape, dtype, device, variant, granularity). Each bound seam previously allocated its own, about 800MB at 24576 tokens over 32 heads, which is why this form had only ever been judged one block at a time — a transformer whose every block reaches the seam ran out of memory before it could be measured. Sharing is sound because nothing survives a call, and the pool is keyed, pointer-stable and never emptied. Across all 48 blocks memory now grows by the quantized weights alone, 0.41 GB per block, with per-block parity unchanged. Naming a form does not force it — the family still qualifies the shape, speed-gates the result, and falls through when the installed package does not serve the site — and an unknown name raises rather than silently returning the default order's result.

## Measured

What a request costs, wall clock, on consumer Blackwell (32GB). The baseline is the unmodified Diffusers host: a 44GB bf16 checkpoint that does not fit on this card, so it runs with weight offloading, which is where a user of this model on this class of hardware actually starts. Same prompt and seed, distilled single-pass recipe, medians of three warm runs, eager.

| Request | Host (offload) | `scheme="nvfp4_balance"` | `scheme="nvfp4_balance_sage"` |
|---|---|---|---|
| 768×512×49f | 99.8 s | 6.0 s (16.6×), peak 29.9 GB | **5.7 s (17.5×)**, peak 26.8 GB |
| 1536×1024×121f | 181.6 s | **87.9 s (2.07×)**, peak 28.0 GB | does not fit yet |

Frames are inspection-equivalent to the host's own output. Attaching is per block, because the bf16 checkpoint is not resident on this card: each block is materialized alone, attached on its own real inputs, and its host weights released before the next.

The full-size row deserves reading closely rather than quoting. All 48 blocks attach, each block's gate measures 1.48×, and the assembled pipeline sits near the card's limit: 23.9 GB resident before the request starts, 28.0 GB at peak, with video-VAE tiling needed for decode to have room. Three things account for the distance between 2.07× and what the same card reaches with a hand-assembled configuration:

- the measurement is eager, with no compilation of the block stack;
- the quantized attention profile does not fit at this size — its per-shape staging pool costs about 3 GB more, and assembly runs out;
- the audio feed-forward stays at host precision throughout, 3.0 GB across the model, because its 126-row calls sit outside the fused chain's 128-row alignment and the seam declines them rather than read an output the kernel never wrote.

### Where the time goes

A diagnostic, not the result: which family earned which part of the request time. Real weights, real captured deployment inputs, paired alternating timing inside the gate.

| Site shape | `scheme=` | Block latency | Attention family | Peak memory |
|---|---|---|---|---|
| S=24576 | host, unattached | 134.3 ms | — | 12.2 GB |
| | `"nvfp4_balance"` | 117.1 ms (1.15×) | BF16 form bound, declined at 1.006× | 8.2 GB |
| | `"nvfp4_balance_sage"` | **89.8 ms (1.49×)** | activated, 1.259× | +3 GB pool |
| S=2688 | host, unattached | 10.2 ms | — | 2.3 GB |
| | `"nvfp4_balance"` | 8.2 ms (1.25×) | declined | 1.7 GB |
| | `"nvfp4_balance_sage"` | 8.0 ms (1.28×) | activated, 1.022× | 4.7 GB |

Matched-forward cosine against the host's own output is 0.99999 in every row; `detach` restores it bit-exactly (max-abs 0.0); ledger fallbacks are zero.

Two numbers in that table are easy to misread. A block ratio is not a kernel ratio: the attention this profile selects measures 45.9 → 19.6 ms (2.34×) on its own, and appears as 1.259× for the attention unit because the unit is judged against the whole block — the same 27 ms, a different denominator. And the 1.006× row is the BF16 form's verdict (46.1 ms against the host's 45.9), not a verdict on the quantized ones.

The sage3 form refuses head_dim 128 on a CUDA 12.8 host, because that is what the published package advertises there — its 128-wide build is CUDA 13 only. The ladder falls through and the refusal is on the trail. This repository keeps no second architecture table.

## Not changed here

No `structure.yaml` bytes, so no `spec_digest` moves and no receipt is orphaned. The `attention_core` variant enum does not name these forms; extending it is a version decision left to maintainers, and `gates.py` currently validates variant keys rather than values, which is worth settling in the same bump. No `csrc`, bindings or CMake. The default attention order, and therefore every other host's behaviour, is unchanged.

### Structures self-review

Scope
- [x] I identified the owning layer and kept unrelated model/kernel work out.
- [x] I listed affected and intentionally unaffected hosts, shapes, dtypes, phases, devices, and public APIs.
- [x] A new catalog boundary has cross-host evidence; otherwise it remains a binding/host concern or is explicitly provisional. — no new catalog boundary; existing `attention_core` and `vision_ffn` are used as published.

Contracts and failure modes
- [x] Discovery is structural, with positive and negative cases.
- [x] No catalog spec bytes changed.
- [x] Hardware capability comes from the kernel package; this change adds no duplicate architecture table.
- [x] Unsupported or missing capability refuses clearly or takes a declared host fallback, and the ledger records it.
- [x] Calibration and precision use the existing collector/scheme entry point.
- [x] Attach/detach, state ownership, repeated calls, and compile/capture behavior are preserved where applicable.

Evidence
- [x] I added focused public tests for the changed contract. — `tests/test_sage_attention_forms.py`, 30 cases.
- [x] The intended path ran (call count > 0) and unexpected fallback count is 0.
- [x] Numerical comparison uses the declared boundary and tolerance.
- [x] Performance claims use paired final-form measurements, not summed microbenchmarks, and include the qualification cell.
- [x] I recorded exact commands and results below.

Maintenance and hygiene
- [x] I checked every changed shared helper/schema reader and every documented API against source. — `bind_dense_attention_best` gained a keyword with a default that preserves every existing call.
- [x] The diff contains no secrets, credentials, private/local paths, checkpoints, generated binaries, logs, or benchmark traces.
- [x] Optional dependencies remain optional at import time; errors name the missing capability and a supported remedy.
- [x] User-facing behavior, support limits, and invalidated performance receipts are documented.

Validation
- Environment: consumer Blackwell (SM120, 32GB), CUDA 12.8 and CUDA 13.2 hosts, torch 2.11 and 2.13, Diffusers 0.40 dev.
- Commands and results: `pytest tests/test_sage_attention_forms.py` 30 passed; with `tests/test_ltx25_contracts.py tests/test_attention_variant_family.py` 73 passed; `tests/test_structures_bindings.py tests/test_structures_attention_logical_dims.py` 35 passed. Bindings validate under `load_binding(require_pipeline_coverage=True)`.
- Correctness/path/ledger: matched-forward cosine 0.999994 (S=24576) and 0.999995 (S=2688) against the host's own output; attention unit band pass at max-abs 1–2; ledger 0 fallbacks over 1602 guarded calls; detach max-abs 0.0.
- Performance: the table above; both arms timed alternately in the gate, same process, eager assembly, medians with spread 0.003–0.007.
- Not run or not covered: the sage3 form's 128-wide path needs a CUDA 13 host and is qualified there (fused entry, all-in 12.30 ms at S=24576, bit-identical to the two-stage path, capture-safe); on the CUDA 12.8 host used for the attach measurements it refuses by envelope. Full-size whole-model attach with quantized attention is blocked on workspace pooling, as recorded above.

### DIFF
--- a/docs/ltx25_structures_design.md
+++ b/docs/ltx25_structures_design.md
@@ -1,0 +1,125 @@
+# LTX-2.5 on the structures layer — design
+
+Scope: promote the LTX-2.5 integration (attention swap, W4A4 FFN chain,
+quantize-on-adopt) from model-private modules into catalog structures, so the
+same regions attach to any host that binds them — the native frontend here,
+and diffusers-hosted checkpoints through a binding, with no model-specific
+code in any impl.
+
+Everything below follows `docs/structures.md`: specs name positions, bindings
+place them on a host, impls decide what runs there. Nothing in this design
+adds a second vocabulary for calibration or qualification.
+
+## 1. What generalizes, and to which structure
+
+| Region (measured on LTX-2.5) | Catalog structure | Status |
+|---|---|---|
+| unmasked self/cross attention, head_dim 128 | `attention_core` | new backend impl |
+| GELU FFN pair (proj → tanh-GELU → down) | `vision_ffn` | new backend impl |
+| per-linear NVFP4 weight adoption | `quantize_on_adopt` | new scheme + binding attribute |
+| adaLN (rms · (1+scale) + shift, per-token tables) | host stage | future `adaln_producer` backend |
+| q/k RMSNorm + RoPE | host stage | future `qk_norm_rope` backend |
+| two-stage denoise pipeline | `video_generation_pipeline` | new pipeline binding |
+
+The audio branch (head_dim 64, short sequences) stays on the host attention
+path by measurement: quantized attention loses to SDPA at those shapes, so
+the binding simply does not claim those sites.
+
+## 2. `attention_core` backend: `sage2_qk_int8_pv_fp8`
+
+Executable form: per-warp INT8 quantization of Q, per-block INT8 of K,
+per-channel FP8 of V, one fused attention kernel, bf16 out. The kernel and
+its quantizers ship as one Hub artifact; the impl reads the supported head
+dims and layouts from the artifact instead of duplicating capability
+knowledge, exactly as the FA2 backend does.
+
+Qualification (all decided from real captures, refusal is legible):
+
+- head_dim must be advertised by the artifact (128 today); other dims return
+  no binding so the host keeps its own attention,
+- masked sites are not claimed — a mask that packs to a dense run can ride
+  the existing packed-KV plan later; today the masked path stays host,
+- scratch (int8/fp8 staging + output) is allocated per shape and shared
+  across all same-shaped sites; call sequences are pointer-stable, so the
+  region is CUDA-graph capturable.
+
+Parity gate: the spec's `real_distribution` rule. Measured on the target
+model this backend holds ~0.9992 cosine per call against an fp32 reference,
+and matched-input single-forward parity sits inside the noise floor of any
+same-precision kernel substitution; the latency rule is satisfied with
+2.0-2.4x over the strongest SDPA backend at the model's sequence lengths.
+
+## 3. `vision_ffn` backend: `w4a4_nvfp4_cutlass`
+
+Executable form, three launches replacing six:
+
+    activation quantize (bf16 -> NVFP4 + block scales)
+    up GEMM with bias + tanh-GELU + NVFP4 output epilogue
+    down GEMM (bf16 out; bias added when the slot carries one)
+
+Weight slots come from the spec; the impl accepts either origin:
+
+- **prequantized hosts** (checkpoint ships NVFP4): dequantize with the
+  reference kernel, requantize into the executable layout at adopt,
+- **bf16 hosts**: direct quantize at adopt (~seconds for a 22B model).
+
+Qualification:
+
+- both dims divisible by 16; rows padded to 128 through a staging buffer when
+  the host batches oddly — the GEMM rejects unaligned M *without writing
+  output*, so the impl owns the pad rather than trusting a return code,
+- adopt is layer-by-layer so peak memory stays near the fp4 footprint,
+- parity is gated against the plain-torch reference on real captures; on the
+  target model the chain holds the same distance from a bf16 golden as the
+  host's own W4A4 path while being 1.25-1.3x faster.
+
+## 4. `quantize_on_adopt`: the site list is a binding attribute
+
+The measured result that shapes this design: blanket adoption of every large
+linear visibly damages output, while adopting exactly the checkpoint
+author's calibrated selection (per-block attention/FFN linears, minus the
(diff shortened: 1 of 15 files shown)


LiangSu8899:
After the structures split (#197), this PR continues in two places: the bindings in https://github.com/flashrt-project/FlashRT/pull/198, the forms and adapters in https://github.com/flashrt-project/FlashRT-Structures/pull/1 (migrated with history). Leaving this open for the record until both land.

PR #179, opened 2026-08-18
tools: Omega-QVLA dit_svdquant_v1 pack → E0M3/UE4M3 converter + format doc + synthetic fixture

Lmy271828:
## Scope

Offline tooling for converting Omega-QVLA `dit_svdquant_v1` rank-0 packs into the packed E0M3 weight and tile-interleaved UE4M3 SFB format consumed by FlashRT's existing SM110 GEMM.

This PR does **not** add a frontend, pipeline, runtime consumer, server, CUDA graph, CMake source, pybind symbol, or public Python API. It has no effect on existing model inference or default builds.

Included:

- `tools/convert_omega_pack_e0m3.py`: schema validation and offline conversion.
- `tools/check_omega_e0m3_layer.py`: CPU emulation plus direct converted-artifact GEMM verification on Thor.
- `tools/gen_omega_pack_fixture.py`: deterministic four-record synthetic fixture.
- `docs/omega_pack_e0m3.md`: Milestone 1 format and reproduction contract.

## Audit follow-up

Commit `7355c3c6` addresses the merge review:

- Removes documentation for runtime/serving/graph files that are not present in this PR. External runtime experiments are no longer represented as repository functionality.
- Validates every selected record before CUDA work: format, rank=0, W4A4 metadata, N/K metadata, low-rank shapes, positive finite scale tables, rotation shapes, finite tensors, and a real int64 permutation.
- Fails instead of skipping unsupported records. An unfiltered complete pack requires 252 records by default; fixtures and filtered subsets require complete selected coverage.
- Adds CPU-only `--validate-only` for schema and coverage checks.
- Records schema version, selected layers, and source/selected counts in the artifact. `weights`, `aux`, and `selected_layers` must cover the same set.
- Writes through a unique sibling temporary file and atomically replaces the destination only after complete conversion. The source pack cannot be used as `--out`.
- Adds `--artifact` to the checker. This path quantizes only the activation and feeds the artifact's stored `packed` and `sfb` tensors directly to `cutlass_fp4_gemm_e0m3w`; it fails when artifact-vs-fp16 cosine is below `--min-artifact-cos`.
- Narrows `.gitignore` from global `*.pt` to `/artifacts/omega_e0m3/*.pt`.
- Adds 13 CPU contract tests.

## Reproduce

CPU validation and emulation:

```bash
python tools/gen_omega_pack_fixture.py --out /tmp/fixture_pack.pt
python tools/convert_omega_pack_e0m3.py \
    --pack /tmp/fixture_pack.pt --out /tmp/unused.pt --validate-only
python tools/check_omega_e0m3_layer.py \
    --pack /tmp/fixture_pack.pt --mode emulate
```

Thor direct artifact round-trip:

```bash
python tools/convert_omega_pack_e0m3.py \
    --pack /tmp/fixture_pack.pt --out /tmp/fixture_e0m3.pt --fold none
python tools/check_omega_e0m3_layer.py \
    --pack /tmp/fixture_pack.pt --artifact /tmp/fixture_e0m3.pt \
    --mode kernel --min-artifact-cos 0.98
```

Full pack conversion defaults to a strict `252/252` coverage gate. Use `--expected-records` only for a deliberately different complete pack.

## Validation

- Tool contract tests: `13 passed`.
- Changed Python scripts: `py_compile` passed.
- `git diff --check`: passed.
- Default fixture generation: passed; four schema records load with `weights_only=True`.
- Fixture CPU validation: `4/4` passed.
- Fixture CPU emulation: passed.
- Private path scan: clean.

The Thor host was offline during this audit follow-up, and the real 252-record pack was not available locally. Before merge, rerun the two hardware gates introduced here: fixture `packed`/SFB direct GEMM round-trip and real-pack `252/252` conversion plus representative direct artifact GEMM checks. Cross-platform model E2E is not required because this PR has no runtime integration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,4 +44,4 @@ internal-tests/
 internal-docs/
 notes/
 third_party/cutlass
-*.pt
+/artifacts/omega_e0m3/*.pt
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ internal-tests/
 internal-docs/
 notes/
 third_party/cutlass
+*.pt
--- a/docs/omega_pack_e0m3.md
+++ b/docs/omega_pack_e0m3.md
@@ -1,0 +1,386 @@
+# Omega-QVLA pack format and the E0M3 consumption contract
+
+## 0. Background concepts (90 seconds)
+
+- **4-bit quantization**: store `round(x / s)` (a small integer) plus the
+  "ruler" `s` (the scale) instead of `x`. Compute happens as
+  `integer × s`. Fewer bits = less memory bandwidth, more rounding error.
+- **fake-quant**: quantize then *immediately dequantize*, staying in float.
+  Simulates quantization error without needing integer hardware — Omega's
+  whole runtime is fake-quant emulation on plain PyTorch matmuls.
+- **scale granularity**: how many elements share one ruler. *Per-channel* =
+  each of the K input channels gets its own (Omega's choice). *Per-16
+  block* = 16 adjacent elements share one (the hardware format's choice).
+  Coarser granularity = fewer scales to store, but elements of different
+  magnitudes get crushed under a shared ruler.
+- **static vs. dynamic scale**: *static* = measured offline on calibration
+  data, stored in the pack (Omega's `act_scale_table`). *Dynamic* =
+  computed per token at runtime from the actual data (`amax / 7`). Dynamic
+  is fresher but constrains the layout to what hardware computes cheaply.
+- **E0M3**: the 4-bit element format here — sign + 3 mantissa-ish bits
+  decoding the uniform integer grid −7..+7. "Uniform" = evenly spaced
+  levels, unlike E2M1 (NVFP4) whose levels bunch near zero.
+- **UE4M3**: an unsigned 4-exponent/3-mantissa mini-float used *only for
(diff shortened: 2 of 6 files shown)


LiangSu8899 (approved):

LiangSu8899:
Audit fixes are pushed in `7355c3c6`.

The PR is now limited to its actual Milestone 1 surface. The converter validates all selected records and fails on unsupported format/rank/shape/rotation/permutation/scale metadata; full-pack conversion defaults to a strict 252-record gate. Output is atomic and records explicit schema/coverage metadata.

The checker now has a real artifact round-trip: `--artifact` loads the converted `packed` and `sfb` tensors and feeds them directly to the SM110 GEMM. It does not regenerate the weight from `weight_res_q`, and it returns failure below `--min-artifact-cos`.

Validation completed locally:

- 13 tool contract tests passed.
- Default four-record fixture generated and loaded with `weights_only=True`.
- CPU validation reported `4/4`.
- CPU emulate checker passed.
- `py_compile` and `git diff --check` passed.
- Merge simulation against current `main` is clean.

The Thor host was offline and the real 252-record pack was not present locally. The PR body now lists the remaining pre-merge hardware gates precisely: direct fixture packed/SFB GEMM round-trip, then real-pack 252/252 conversion and representative direct artifact GEMM checks. No model E2E is needed for this offline-only PR.

PR #181, opened 2026-08-20
fix(kernels): rename misnamed gate_silu_mul* to gate_geglu*

hebangwen:
## Summary

This PR fixes a naming error in `csrc/kernels/activation.cu` where a family of kernels was misleadingly named `gate_silu_mul*`, even though they compute the **tanh/sigmoid-approximated GELU** (GeGLU, i.e. `GELU(gate) * up`), not SiLU.

The incorrect naming has been noted in two places already:

- `csrc/kernels/act_fuse_sm120.cuh` — *"the existing `gate_silu_mul` binding is misnamed -- it is the tanh-approx GELU, not SiLU"*
- `csrc/kernels/silu_mul_qwen36.cuh` — *"the existing `gate_silu_mul` ... actually computes GELU(gate)*up ... This kernel is the proper SiLU variant"*

This PR renames the misnamed symbols to reflect their true semantics, so the C++ symbol names finally match the `gate_geglu` naming already exposed to Python via the bindings.

## Changes

- **Rename-only, no numerical/behavioral changes** — activation math is untouched.

| Symbol (before → after) |
| --- |
| `gate_silu_mul` → `gate_geglu` |
| `gate_silu_mul_fp16` → `gate_geglu_fp16` |
| `gate_silu_mul_merged` → `gate_geglu_merged` |
| `gate_silu_mul_merged_fp16` → `gate_geglu_merged_fp16` |
| `gate_silu_mul_merged_fp8` → `gate_geglu_merged_fp8` |
| `gate_silu_mul_merged_fp8_fp16` → `gate_geglu_merged_fp8_fp16` |

Files touched:

- `csrc/kernels/activation.cu` — kernel + wrapper renames
- `csrc/kernels/activation.cuh` — declaration renames
- `csrc/bindings.cpp` — internal call-site renames (Python-facing names already `gate_geglu*`)
- `cpp/models/pi05/src/targets/sm110/operation_driver.cu` — 2 call sites
- `cpp/models/pi05/src/targets/sm120/target.cpp` — 3 call sites

## Not touched

The genuinely-SiLU kernels (`silu_mul_split_fp8_kernel`, `silu_inplace_kernel`, `fused_add_silu_kernel`) are correct and left unchanged.

## Why rename instead of changing the math

Changing the GELU formula to SiLU would silently break every caller that relies on GeGLU semantics (FFN intermediate layers, Pi0/VLA encoder paths). Renaming preserves behavior while fixing the semantic mismatch.

### DIFF
--- a/cpp/models/pi05/src/targets/sm110/operation_driver.cu
+++ b/cpp/models/pi05/src/targets/sm110/operation_driver.cu
@@ -358,7 +358,7 @@ modalities::Status Sm110OperationDriver::gate_gelu_fp16(
     if (!merged || !output || rows <= 0 || hidden <= 0) {
         return invalid("SM110 FP16 gated GELU arguments are invalid");
     }
-    ::gate_silu_mul_merged_fp16(
+    ::gate_geglu_merged_fp16(
         static_cast<const __half*>(merged), static_cast<__half*>(output),
         rows, hidden, reinterpret_cast<cudaStream_t>(stream));
     return launch_status();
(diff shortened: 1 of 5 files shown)


LiangSu8899 (approved):

LiangSu8899:
Thanks for fixing!

PR #189, opened 2026-08-30
AMD MI350X (CDNA4) backend — Pi0.5 at 16.4 ms FP8

LiangSu8899:
## Summary

Adds a self-contained AMD backend targeting Instinct MI350X (CDNA4 / gfx950, ROCm 7.x), with Pi0.5 as the first supported model. The port keeps every FlashRT contract: bare-kernel hot path (zero Python/framework ops at inference), raw-pointer ABI, static-graph capture-then-replay, real-data FP8 calibration.

The NVIDIA tree is untouched except for two documented extension points (`detect_arch` / pipeline map) and an arch-aware extension check in `api.py`.

## Performance

Pi0.5 (2 camera views, 10-step denoise), median E2E latency on real LIBERO frames, MI350X:

| Configuration | Latency | vs torch.compile |
|---|---|---|
| PyTorch eager | 116.9 ms | — |
| torch.compile (max-autotune) | 40.5 ms | 1.00× |
| **FlashRT AMD, FP8 (default)** | **16.4 ms** | **2.47×** |
| FlashRT AMD, BF16 | ~30 ms | 1.34× |

Same protocol on an RTX 5090 (FlashRT FP8, CUDA) measures 19.7 ms — this port is currently the fastest FlashRT Pi0.5 target.

Accuracy: output cosine vs the FP32 openpi reference is 0.9992–0.9994 (FP8, noise pinned to the reference draw), replay is bit-stable within a process. Fixed state-prompt mode costs ~0.1–0.3 ms over exact mode at a right-sized capacity and serves changing prompt lengths without recapture.

## Usage

- **`docs/deployment_amd.md`** — build, precision tiers (FP8 default / BF16 escape / FP4 status), prompt-length graph strategies (`exact` / `fixed` + `state_prompt_fixed_max_len`), environment knobs, feature matrix vs the RTX frontend, reproduction protocol, HIP-vs-CUDA porting notes
- **`examples/pi05_amd_quickstart.py`** — runnable stable-door quickstart (`load_model(..., hardware="amd_cdna4")` is auto-detected on ROCm)
- **`csrc/amd/README.md`** — standalone HIP build (`scripts/amd/build_amd.sh`), no third-party dependencies; hipBLASLt ships with ROCm

## Layout

- `csrc/amd/` — standalone CMake project (does not touch the root CUDA build): pybind module with the same pointer ABI, hipBLASLt GemmRunner (+ timed autotune), hand-written CDNA4 kernels — MFMA packed-weight small-M FP8 GEMMs, split-KV decoder attention (FA2-style `seqused`, fused FP8-out epilogue), MFMA encoder flash attention, norm/rope/quantize/fusion families tuned for gfx950
- `flash_rt/amd/` — ctypes HIP runtime twins (`hip_buffer` / `hip_graph`), CDNA4 attention backends (aiter with sdpa fallback), Pi0.5 pipeline + torch frontend mirroring the RTX feature surface (calibration single/multi-frame, precision spec, temporal K/V caching, fixed/exact state-prompt modes)

## Validation

All validation is in the repo suite below. Full-graph capture is a hard gate; replay outputs are verified bit-stable; the end-to-end gate judges against a saved FP32 reference with the denoise noise pinned, reporting medians against dual baselines (eager and torch.compile).

### Running the tests

On a gfx950 machine with ROCm 7.x and a ROCm PyTorch build:

```bash
# 1. build the extension (rejects a non-gfx950 target)
bash scripts/amd/build_amd.sh gfx950

# 2. hardware-independent gates — also runnable on a CUDA-only box,
#    where every ROCm-dependent case skips with a stated reason
python -m pytest tests/test_amd_routing.py tests/test_amd_arch_gate.py -v

# 3. extension, HIP graph and kernel-parity gates (needs the built .so
#    and a ROCm device)
python -m pytest tests/test_amd_extension.py tests/test_amd_hip_graph.py \
                 tests/test_amd_kernel_parity.py -v

# 4. model-level gates (needs a Pi0.5 checkpoint; the reference-cosine
#    case additionally needs a saved actions .npy)
export FLASH_RT_PI05_AMD_CKPT=<pi05_checkpoint_dir>
export FLASH_RT_PI05_AMD_REF_ACTIONS=<reference_actions.npy>   # optional
python -m pytest tests/test_amd_pi05_model.py -v

# 5. end-to-end smoke with timing
python examples/pi05_amd_quickstart.py --checkpoint <pi05_checkpoint_dir>
```

Results on an MI350X: **50 passed, 2 skipped, 0 failed** (steps 2–4; the two skips are the optional reference-cosine case and the missing-extension case, which cannot fire once the extension is built). On a CUDA-only machine the same suite collects cleanly and reports **13 passed, 39 skipped**.

Note on the environment: attention runs on aiter when it is importable and falls back to torch SDPA otherwise, which costs about 6 ms end-to-end here (16.3 ms versus 22.2 ms, one variable changed in one process). The numbers above are with aiter available.

### DIFF
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -43,6 +43,7 @@ set(CMAKE_HIP_ARCHITECTURES "${GPU_ARCH}")
 set(CMAKE_HIP_STANDARD 17)
 
 find_package(hip REQUIRED)
+find_package(hipblaslt REQUIRED)
 
 # ── pybind11 (same resolution as root CMake: ask the interpreter) ──
 find_package(Python COMPONENTS Interpreter Development REQUIRED)
@@ -56,6 +57,11 @@ find_package(pybind11 CONFIG REQUIRED)
 set(FLASHRT_AMD_SOURCES
   bindings.cpp
   kernels/norm.hip
+  kernels/activation.hip
+  kernels/elementwise.hip
+  kernels/rope.hip
+  kernels/patch_embed.hip
+  gemm/hipblaslt_runner.hip
 )
 set_source_files_properties(${FLASHRT_AMD_SOURCES} PROPERTIES LANGUAGE HIP)
 
@@ -65,7 +71,7 @@ target_compile_definitions(flash_rt_amd_kernels PRIVATE
   FLASHRT_AMD_GPU_ARCH="${GPU_ARCH}")
 target_compile_options(flash_rt_amd_kernels PRIVATE
   $<$<COMPILE_LANGUAGE:HIP>:-O3 -ffp-contract=fast>)
-target_link_libraries(flash_rt_amd_kernels PRIVATE hip::host)
+target_link_libraries(flash_rt_amd_kernels PRIVATE hip::host roc::hipblaslt)
 
 set_target_properties(flash_rt_amd_kernels PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/../../flash_rt/amd")
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -63,7 +63,9 @@ set(FLASHRT_AMD_SOURCES
   kernels/patch_embed.hip
   kernels/fusion.hip
   kernels/quantize_fp8.hip
+  attention/decoder_flash.hip
   gemm/hipblaslt_runner.hip
+  gemm/smallm_fp8.hip
 )
 set_source_files_properties(${FLASHRT_AMD_SOURCES} PROPERTIES LANGUAGE HIP)
 
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -61,6 +61,8 @@ set(FLASHRT_AMD_SOURCES
   kernels/elementwise.hip
   kernels/rope.hip
   kernels/patch_embed.hip
+  kernels/fusion.hip
+  kernels/quantize_fp8.hip
   gemm/hipblaslt_runner.hip
 )
 set_source_files_properties(${FLASHRT_AMD_SOURCES} PROPERTIES LANGUAGE HIP)
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -67,6 +67,7 @@ set(FLASHRT_AMD_SOURCES
   kernels/ew_tune.hip
   attention/decoder_flash.hip
   attention/attn_probe.hip
+  attention/encoder_flash.hip
   gemm/hipblaslt_runner.hip
   gemm/smallm_fp8.hip
   gemm/smallm_mfma.hip
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -66,6 +66,7 @@ set(FLASHRT_AMD_SOURCES
   kernels/stream_probe.hip
   kernels/ew_tune.hip
   attention/decoder_flash.hip
+  attention/attn_probe.hip
   gemm/hipblaslt_runner.hip
   gemm/smallm_fp8.hip
   gemm/smallm_mfma.hip
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -64,9 +64,11 @@ set(FLASHRT_AMD_SOURCES
   kernels/fusion.hip
   kernels/quantize_fp8.hip
   kernels/stream_probe.hip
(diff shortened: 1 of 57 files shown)


LiangSu8899:
Pushed review fixes (3616e212). Addressing the four blocking items:

**1. HIP errors silently ignored** — every HIP runtime call in the AMD Python layer now routes through the return-code checker and raises `RuntimeError` on failure: the inference D2D copy and stream sync, the calibration syncs, the graph warmup/capture syncs and post-replay syncs, and the `HipBuffer` memset/copy/sync paths. A failed launch, copy or sync can no longer let a stale output buffer be read back as an action. The one deliberate exception is `hipFree` in `__del__`, which logs instead of raising (exceptions in `__del__` are unraisable in Python, and a failed free cannot corrupt a result).

**2. No committed tests** — added five suites (52 tests):
- `tests/test_amd_routing.py` — pipeline-map entry, `detect_arch()` ROCm branch including non-gfx950 rejection and the gfx950-prefix trap, `load_model` failure modes (missing extension names the build command; Thor-only flags rejected)
- `tests/test_amd_extension.py` — `build_info()`/`device_arch()` coherence and an explicit required-symbol inventory of every bound kernel
- `tests/test_amd_hip_graph.py` — buffer round-trips with pattern data, capture → instantiate → replay, byte-identical repeat replays, replay-before-capture refusal
- `tests/test_amd_kernel_parity.py` — kernel-surface parity vs torch on real-distribution inputs, FP8 byte-exactness against a same-scale torch cast, and the seqused fixed-shape attention path with poisoned padding rows
- `tests/test_amd_pi05_model.py` — end-to-end load, graph capture, pinned-noise bit-identity, exact/fixed prompt modes (fixed serves a second length without rebuilding the pipeline or re-capturing), FP8 and BF16 arms, optional reference-cosine gate

Everything skips with a stated reason off ROCm. On an MI350X: **50 passed, 2 skipped, 0 failed**.

**3. Explicit AMD routing did not verify gfx950** — now enforced twice. `scripts/amd/build_amd.sh` rejects a non-gfx950 `GPU_ARCH` (`FLASHRT_AMD_ALLOW_ARCH=1` overrides for a future port), and the frontend's first action in `__init__` — before the checkpoint is touched — compares the extension's `device_arch()` and `build_info()["gpu_arch"]` and raises unless both are gfx950. Forcing `hardware="amd_cdna4"` on another AMD card now fails there instead of running gfx950-only MFMA paths.

**4. Quickstart `--num-views=3` was broken** — the example now builds one synthetic image per configured view and constrains the flag to `{2,3}`. The frontend enforces the same range and validates the observation per view (key present, `(224,224,3)`, `uint8`), so a missing view raises instead of silently reusing stale image buffers.

**Docs** — `docs/deployment_amd.md` is now the model-independent guide (hardware/build/layout/routing/shared knobs/tests/measurement discipline) and `docs/deployment_amd_pi05.md` carries the Pi0.5 specifics. One quantified addition worth flagging for deployments: attention is the largest kernel bucket here, so running without aiter (torch SDPA fallback) costs about 6 ms end-to-end — 16.3 ms vs 22.2 ms, measured with a single variable changed in one process.

**Performance after the fixes is unchanged**: the documented protocol re-measured on the fixed tree gives 16.42 ms median (fixed-shape arms 17.14 / 16.74), matching the published numbers.

LiangSu8899:
Both remaining items are addressed.

**[P1] gfx950 prefix over-acceptance** (46a8b2ee) — correct catch, `startswith("gfx950")` admitted `gfx9500`. All three gates now compare the base target exactly: the frontend takes `arch.split(":", 1)[0] == "gfx950"` (so a `gfx950:sramecc+:xnack-` feature suffix still passes) and the build script compares `${GPU_ARCH%%:*}` against `gfx950`. New `tests/test_amd_arch_gate.py` pins the rule with the `gfx9500` trap on both the device and build sides, asserts neither frontend has regressed to a prefix test, and runs the build script against `gfx9500`/`gfx942`/`gfx90a`/`sm_120` to confirm each is rejected — 15 passed, 1 skipped on a CUDA-only box (the skip is the GROOT frontend, which lives on the stacked branch).

**Description: missing reproducible commands** — the description now carries the full copy-pasteable sequence (build, hardware-independent gates, extension/graph/parity gates, model-level gates with their environment variables, quickstart) together with the results on both an MI350X and a CUDA-only machine.

One correction to an earlier claim of mine while re-verifying for this round: a missing ROCm runtime used to surface as `OSError` from the ctypes layer, which bypassed the loader's ImportError-based hardware gate (users saw a raw shared-library error instead of the build instructions) and broke test collection on CUDA-only machines. Both ctypes layers now re-raise as `ImportError` naming the library (bd5b7b50), which is what made the suite collect cleanly off ROCm.

PR #190, opened 2026-08-30
GROOT N1.7 on AMD MI350X (CDNA4) — 16.0 ms full-frame, 4.24× over eager

LiangSu8899:
Stacked on #189 (the AMD CDNA4 backend). Review that one first — this branch contains its commits.

## Summary

Adds GROOT N1.7 to the AMD MI350X (gfx950 / CDNA4) backend: an FP8 kernel backbone with an unquantized bf16 DiT action head. Same contracts as the rest of the backend — bare-kernel hot path, raw-pointer ABI, capture-then-replay, real-data calibration.

Precision scope, stated precisely: the Thor and RTX FP8 frontends inherit `_DIT_USE_FP8 = True` and additionally quantize the DiT FFN and fused self-attention QKV. This backend sets `_DIT_USE_FP8 = False`, so its DiT is fully bf16 — a **less** quantized tier than its CUDA counterparts. Porting that DiT calibration path to CDNA4 is open work and remains available headroom rather than something this PR ships.

## Performance

GROOT N1.7 (3.1 B), 2 camera views, 4-step flow matching, action horizon 40. Median **full-frame** latency — a fresh observation through the backbone graph plus the action chain, i.e. what a serving loop pays per frame:

| Configuration | Latency | vs eager |
|---|---|---|
| PyTorch eager (official policy) | 67.9 ms | 1.00× |
| torch.compile (max-autotune) | 77.9 ms | 0.87× |
| **FlashRT AMD, FP8 backbone + bf16 DiT** | **16.0 ms** | **4.24×** |

`torch.compile` is slower than eager for this model on this stack, so the judged baseline is eager. The frame splits into roughly 7 ms of backbone replay and 9 ms of action chain; deployments that reuse backbone features across frames pay only the latter.

Accuracy: combined denormalized-action cosine 0.9995 against the reference with the initial noise pinned (the 1-D gripper channel is near-constant over a trajectory, so its cosine is not a useful gate and is reported separately).

How it got there, each step an A/B with one variable changed in one process and an accuracy gate:

| Step | Full-frame median |
|---|---|
| First working port | 21.2 ms |
| Timed hipBLASLt algorithm selection | 20.6 ms |
| Fused GEMM epilogues (backbone + DiT) | 18.4 ms |
| Fused frontend closure epilogues | 18.2 ms |
| Fused norm→quantize chains, residual-accumulate GEMMs | 17.5 ms |
| Packed-weight MFMA GEMMs on the DiT projections | 16.0 ms |

## What is new here

- `flash_rt/amd/models/groot_n17/pipeline.py` — the five backbone forwards plus an AMD-local `dit_forward`. The decomposed form mirrors the RTX FP8 path exactly; the fused tier uses hipBLASLt epilogues that the CUDA SM120 path cannot (fused FP8 bias and bias+GELU), so this backend starts from a more fused shape than its CUDA twin.
- `flash_rt/amd/frontends/torch/groot_n17.py` — subclasses the existing GROOT frontend and its FP8 backbone mixin, overriding eight hooks (kernel-module seeding, the backbone stage driver, the attention backend, cross-KV setup and refresh, DiT graph capture, and the processor load). `torch.cuda.CUDAGraph` is HIP-backed, so the base class's graph machinery is reused unchanged.
- `flash_rt/amd/hardware/cdna4/attn_backend_groot_n17.py` — one backend for all five attention sites. aiter beat torch SDPA at every measured GROOT geometry (1.19×–1.83×), including head_dim 48 and the GQA causal site; `FVK_AMD_ATTN=sdpa` remains the fallback. The LLM K/V slots keep the model's native 8 KV heads because aiter does GQA internally.
- `csrc/amd/gemm/` — FP16 GEMM and FP8 bias/GELU/descale epilogue entry points (signature-identical to the CUDA class), lazy timed algorithm selection, and a bf16 packed-weight small-M MFMA family.
- `csrc/amd/kernels/` — the FP16 elementwise / norm / rope / quantize families and LayerNorm-style AdaLN kernels the model needs.

## Kernel routing policy

The hand-written MFMA GEMM is routed only where it measured faster than the autotuned library: the DiT projections at M=41 (1.41× on the bias epilogue, 951 GB/s versus 673 GB/s effective weight bandwidth). The FFN shapes measured slower and deliberately stay on hipBLASLt. `FVK_AMD_DIT_GEMM=hipblaslt` disables the routing; `FVK_AMD_FUSED_EPILOGUE=0` selects the decomposed form. Both are read at setup only.

## Validation

Every entry point this branch adds to `csrc/amd` is gated numerically. The three binding fragments contribute 54 entry points (21 in the hipBLASLt runner, 4 in the packed-weight MFMA family, 29 in the fp16/bf16/fp8-fusion/AdaLN kernels) and all 54 now have tests:

- `tests/test_amd_kernel_parity.py` — extended to the full `GemmRunner` surface (`fp16_nn`, `fp8_nn_bias`, `fp8_nn_gelu_bias`, `fp8_descale_fp16`, `bf16_nn_res`, `bf16_nn_bias`, `bf16_nn_bias_gelu`, `bf16_nn_bias_res`, `bf16_run`, `mxfp4_nt_dev`, and the autotune entry points).
- `tests/test_amd_kernel_parity_groot.py` — the packed-weight `smallm_mfma_bf16_*` family and the 29 hand-written kernels, including `ada_layer_norm_fp8`, `silu_mul_split_fp8_fp16`, `layer_norm_fp8_static_fp16_vec` and the `*_vec` status contracts.
- `tests/test_amd_groot_routing.py` — pipeline-map entry, `load_model` precision-tier contracts, attention-backend site and layer-index validation. Runs without a GPU.
- `tests/test_amd_groot_model.py` — end-to-end: `set_prompt` through the FP8 kernel backbone (asserted via the baked per-stage scales, so a silent fallback fails the test), finite actions of the expected shape, bit-identical results under a pinned initial noise, and a cosine gate against a reference fixture when provided.

The gates assert semantics rather than a loose cosine: FP8 outputs are byte-compared against a `torch.float8_e4m3fn` cast through the same device scale, replicating the kernel's reciprocal-multiply (zero permitted byte differences on the pure multiply-clamp-convert paths); residual-accumulate variants run against a pre-seeded destination and must not match the non-accumulated form; bias and GELU variants are pinned to the tanh approximation with a discriminator that fails if the term is dropped; and the packed-weight GEMM is additionally fed the same weights unpacked, which must produce a wrong answer, proving the permutation is load-bearing.

### Running the tests

On a gfx950 machine with ROCm 7.x and a ROCm PyTorch build:

```bash
# 1. build the extension (rejects a non-gfx950 target)
bash scripts/amd/build_amd.sh gfx950

# 2. hardware-independent gates — also runnable on a CUDA-only box,
#    where every ROCm-dependent case skips with a stated reason
python -m pytest tests/test_amd_routing.py tests/test_amd_arch_gate.py \
                 tests/test_amd_groot_routing.py -v

# 3. per-kernel parity for every entry point (needs the built .so
#    and a ROCm device)
python -m pytest tests/test_amd_kernel_parity.py \
                 tests/test_amd_kernel_parity_groot.py -v

# 4. extension inventory and HIP graph capture/replay
python -m pytest tests/test_amd_extension.py tests/test_amd_hip_graph.py -v

# 5. model-level gates (checkpoints and fixtures via environment)
export FLASH_RT_GROOT_N17_AMD_CKPT=<groot_n17_checkpoint_dir>
export FLASH_RT_GROOT_N17_AMD_AUX=<processor_aux_bundle.pt>
export FLASH_RT_GROOT_N17_AMD_REF=<reference_fixture.pt>        # optional
export FLASH_RT_PI05_AMD_CKPT=<pi05_checkpoint_dir>             # for the pi05 gates
python -m pytest tests/test_amd_pi05_model.py \
                 tests/test_amd_groot_model.py -v
```

Results on an MI350X: step 3 gives **111 passed, 1 skipped** (the skip is `mxfp4_nt_dev`, which hipBLASLt reports no algorithm for at that shape — a library capability gap, not a kernel regression), and steps 2, 4 and 5 together give **90 passed, 2 skipped**. Total **201 passed, 3 skipped, 0 failed**. On a CUDA-only machine the whole suite collects cleanly and every hardware-dependent case skips with a stated reason.

Attention runs on aiter when importable and falls back to torch SDPA otherwise; the numbers above are with aiter available.

## Docs

`docs/deployment_amd_groot_n17.md` (model guide: performance, precision tier, attention, fusion and routing knobs, tests, reproduction) linked from `docs/deployment_amd.md`, which is the model-independent AMD guide.

### DIFF
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -63,6 +63,9 @@ set(FLASHRT_AMD_SOURCES
   kernels/patch_embed.hip
   kernels/fusion.hip
   kernels/quantize_fp8.hip
+  kernels/elementwise_fp16.hip
+  kernels/norm_fp16.hip
+  kernels/adaln_layer_norm.hip
   kernels/stream_probe.hip
   kernels/ew_tune.hip
   attention/decoder_flash.hip
--- a/csrc/amd/CMakeLists.txt
+++ b/csrc/amd/CMakeLists.txt
@@ -74,6 +74,7 @@ set(FLASHRT_AMD_SOURCES
   gemm/hipblaslt_runner.hip
   gemm/smallm_fp8.hip
   gemm/smallm_mfma.hip
+  gemm/smallm_mfma_bf16.hip
   gemm/decoder_ffn_fused.hip
 )
 set_source_files_properties(${FLASHRT_AMD_SOURCES} PROPERTIES LANGUAGE HIP)
--- a/csrc/amd/bindings.cpp
+++ b/csrc/amd/bindings.cpp
@@ -27,6 +27,7 @@
 #include "gemm/decoder_ffn_fused.h"
 #include "kernels/stream_probe.h"
 #include "kernels/ew_tune.h"
+#include "kernels/fp16_ops.h"
 
 namespace py = pybind11;
 
@@ -690,4 +691,8 @@ PYBIND11_MODULE(flash_rt_amd_kernels, m) {
 
     // ── GEMM: fused decoder-FFN pair (gate|up+geglu, down+gate*res) ──
 #include "gemm/bindings_ffn_fused.inc"
+
+    // ── FP16 backbone port: elementwise/norm/rope/quantize + AdaLN
+    //    (see kernels/{elementwise_fp16,norm_fp16,adaln_layer_norm}.hip) ──
+#include "kernels/bindings_fp16_port.inc"
 }
--- a/csrc/amd/bindings.cpp
+++ b/csrc/amd/bindings.cpp
@@ -24,6 +24,7 @@
 #include "gemm/hipblaslt_runner.h"
 #include "gemm/smallm_fp8.h"
 #include "gemm/smallm_mfma.h"
+#include "gemm/smallm_mfma_bf16.h"
 #include "gemm/decoder_ffn_fused.h"
 #include "kernels/stream_probe.h"
 #include "kernels/ew_tune.h"
@@ -688,6 +689,9 @@ PYBIND11_MODULE(flash_rt_amd_kernels, m) {
             v.append(smallm_mfma_variant_name(i));
         return v;
     });
+
+    // ── MFMA small-M BF16 packed GEMM (see gemm/smallm_mfma_bf16.h) ──
+#include "gemm/bindings_smallm_bf16.inc"
 
     // ── GEMM: fused decoder-FFN pair (gate|up+geglu, down+gate*res) ──
 #include "gemm/bindings_ffn_fused.inc"
--- a/csrc/amd/gemm/bindings_gemm.inc
+++ b/csrc/amd/gemm/bindings_gemm.inc
@@ -22,7 +22,8 @@
 //
 // Notes vs the CUDA bindings:
 //   - Only the ported subset is exposed: FvkContext, bf16_run,
-//     bf16_nn, bf16_nn_bias, bf16_nn_bias_gelu, fp8_nn_dev,
+//     bf16_nn, bf16_nn_res, bf16_nn_bias, bf16_nn_bias_gelu,
+//     bf16_nn_bias_res, fp8_nn_dev,
 //     fp8_nt_dev, autotune_bf16_nn, autotune_fp8_nn_dev,
 //     autotune_fp8_nt_dev.
 //   - AMD-only additions (no CUDA counterpart): mxfp4_nt_dev and
@@ -59,6 +60,12 @@
             self.bf16_nn(to_ptr(A), to_ptr(B), to_ptr(D), M, N, K, to_stream(stream));
         }, py::arg("A"), py::arg("B"), py::arg("D"),
            py::arg("M"), py::arg("N"), py::arg("K"), py::arg("stream") = 0)
+        .def("bf16_nn_res", [](GemmRunner& self,
+                                uintptr_t A, uintptr_t B, uintptr_t D,
+                                int M, int N, int K, uintptr_t stream) {
+            self.bf16_nn_res(to_ptr(A), to_ptr(B), to_ptr(D), M, N, K, to_stream(stream));
+        }, py::arg("A"), py::arg("B"), py::arg("D"),
+           py::arg("M"), py::arg("N"), py::arg("K"), py::arg("stream") = 0)
         .def("bf16_nn_bias", [](GemmRunner& self,
                                  uintptr_t A, uintptr_t B, uintptr_t D, uintptr_t bias,
                                  int M, int N, int K, uintptr_t stream) {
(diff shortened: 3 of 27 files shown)


LiangSu8899:
All three blockers and the description issue are addressed (aee42670, plus 46a8b2ee merged up from #189).

**[P1] Import-time global monkeypatch of `huggingface_hub.model_info`** — you are right, and it should never have reached the frontend; that pattern came from a throwaway benchmark harness where the process is disposable. The module-level patch is gone. The workaround is now a `contextlib.contextmanager` that installs the substitute for the duration of one processor load and restores it in `finally`, applied by an override of `_hf_processor` in the AMD frontend. Nothing else in the process ever sees a modified `model_info`, genuine offline errors elsewhere are untouched, and any exception other than `OfflineModeIsEnabled` still propagates inside the scope.

**[P1] Claimed per-kernel parity with no such tests in the repo** — also correct, and it was the same failure the first review caught on #189: the parity harnesses existed in a development workspace rather than in the PR. They are now in the repo, and I widened the scope while porting them. I enumerated every entry point the branch adds from `git diff origin/main...HEAD` across the three binding fragments — 54 in total — and all 54 are gated. That included twelve you had not named (`bf16_run`, `bf16_nn_res`, `bf16_nn_bias`, `bf16_nn_bias_gelu`, `mxfp4_nt_dev`, `enable_lazy_autotune` and the six `autotune_*` entries), none of which had numerical coverage anywhere. The gates assert dtype, rounding, bias and residual semantics rather than a loose cosine — details in the updated description. Porting also exposed that `test_amd_extension.py`'s required-symbol inventory had drifted and was missing nine `GemmRunner` methods; that list is now complete, which is what that test exists to catch.

**[P1] gfx950 prefix over-acceptance** — fixed in #189 and merged here. All three gates compare the base target exactly (`arch.split(":", 1)[0] == "gfx950"`, and `${GPU_ARCH%%:*}` in the build script), so a feature suffix still passes and `gfx9500` does not. `tests/test_amd_arch_gate.py` pins the rule on both the device and build sides, asserts neither frontend has regressed to a prefix test, and runs the build script against `gfx9500`/`gfx942`/`gfx90a`/`sm_120`.

**Description: "mirroring RTX FP8 production tier"** — that was wrong and I have corrected it in the PR description, the frontend module and class docstrings, and `docs/deployment_amd_groot_n17.md`. The Thor base sets `_DIT_USE_FP8 = True` and the RTX FP8 frontend does not override it, so the RTX DiT FFN and fused self-attention QKV run FP8. This backend sets it to `False`, so its DiT is fully bf16 — a less quantized tier, not a mirror. The upstream docstring's "bf16 DiT" refers to the unquantized dtype choice (bf16 rather than fp16), which is what misled me. The text now says so explicitly and marks the DiT FP8 path as unported headroom.

**Reproducible commands** — both descriptions now carry the full copy-pasteable sequence with the environment variables and the results.

Hardware results after the changes: **201 passed, 3 skipped, 0 failed** on an MI350X (per-kernel parity 111 passed / 1 skipped; the rest of the suite 90 passed / 2 skipped).

LiangSu8899:
Fixed in a57d798e, by removing the substitution rather than making it safe.

**Why not a lock.** You are right that save/restore is not concurrency-safe, and a mutex would only narrow the window rather than remove the class of problem — the substitute is still visible to every other thread while it is installed. I checked whether the lookup could be handled at the transformers call boundary instead, as you suggested: in `tokenization_utils_base._patch_mistral_regex` the guard is `if _is_local or is_base_mistral(...)`, so the metadata lookup is skipped only when the tokenizer resolves from a local path, and no keyword argument (`fix_mistral_regex` included) suppresses it. There is no local interception point, which to me argues the library should not be doing this at all.

So the workaround is gone — no assignment to `huggingface_hub.model_info` at import or at runtime. In its place:

- `_hf_processor` catches `OfflineModeIsEnabled` and raises a `RuntimeError` naming the cause (a transformers 4.57.x issue, not a missing file) and the two remedies.
- `set_hf_processor()` lets a caller build the processor in their own setup code — where a single-threaded script may legitimately do whatever it likes with its own process — and inject it before inference.
- `tests/test_amd_no_global_patching.py` pins the contract three ways: importing either AMD frontend leaves the watched third-party callables identical; the frontend sources contain no assignment to them (so a future workaround hidden inside a function is caught too, not just an import-time one); and an injected processor is used without attempting a load.

**One thing the removal established.** With the workaround gone, the model-level reference-cosine gate — which runs the full `infer` → `denormalize_action` path and therefore loads the processor — still passes under `HF_HUB_OFFLINE=1` on the MI350X: the previous run had 90 passed / 2 skipped and this one has 95 passed / 2 skipped, the five new tests being the difference and nothing having flipped to skipped. The frontend loads the processor from the local checkpoint directory, which takes the `_is_local` short-circuit; the failure I had originally worked around came from the gr00t policy's own `build_processor`, in a benchmark script, not from this code path. The workaround was unnecessary here in the first place.

Hardware results on the current head: per-kernel parity **111 passed, 1 skipped**; the rest of the suite **95 passed, 2 skipped**; **206 passed, 3 skipped, 0 failed** in total.

PR #196, opened 2026-09-10
Ascend NPU backend for Pi0.5 (910B4, BF16 + opt-in static INT8)

LiangSu8899:
Adds an Ascend NPU backend for Pi0.5, in its own subtree, with a BF16 tier and an
opt-in static-INT8 tier.

## Scope

| | |
|---|---|
| part | **Ascend910B4 only** — 20 cube / 40 vector cores, CANN 8.5.2 |
| stack | PyTorch 2.7.1 with matching torch_npu 2.7.1.post2 |
| model | Pi0.5, `framework="torch"`, 2 or 3 camera views |
| tiers | `bf16` (default) and `int8` (opt-in, scales frozen at setup) |

The part is named, not inferred: the host tiling names it and several kernels
divide work by its twenty cube cores, so the build refuses any other target
rather than emitting libraries whose tiling would be wrong. Supporting a second
part means validating it, not relaxing a check.

## Using it

```python
from flash_rt.api import load_model

model = load_model(checkpoint_dir, hardware="npu", precision="int8", num_views=2)
model.calibrate(real_observations, percentile=90)     # INT8 only
actions = model.predict(images, prompt=task, state=normalized_state)
```

`precision` takes `"auto"`, `"bf16"`, `"int8"` or `"fp8"`, and **this backend is
the only one that interprets it**: on every other arch anything but `"auto"`
raises, naming the flag that does select a tier there. Those backends choose
their frontend class from `use_fp8`/`use_fp4`/`use_fp16` several hundred lines
before a generic tier could be consulted, so honouring `precision=` there would
mean something other than what it says — `precision="bf16"` would still land on
an FP4 frontend. Making it the real selector everywhere means normalising the
flags ahead of all routing and testing the full matrix; that is a change to
those backends, not part of adding this one.

`precision="fp8"` raises on Ascend: 910/A2 parts have no FP8 tensor hardware.
The tier is resolved before any backend import, so an impossible request fails
identically on a machine with no CANN runtime installed.

The INT8 tier reads its ranges from what it is shown, so `predict()` raises until
`calibrate()` has run. Calibration goes through the repo's existing unified
`VLAModel.calibrate(observations, percentile=, max_samples=, verbose=)`.

## Build

```bash
bash scripts/npu/build.sh
```

Standalone CCE compilation: no root CMake target, no PyTorch C++ extension, no
entry into the CUDA or AMD build trees. It produces four shared objects, split
because the CANN kernel headers define a per-translation-unit tiling symbol and a
cube-only unit cannot share a file scope with a mixed cube/vector one.

Each library exports its SoC target and an ABI number; every loader verifies both
against each other and against the running device before binding an entry point.
A partial rebuild or a mismatched per-library override therefore fails at load,
naming the library, instead of calling a stale contract behind identical symbol
names.

## Fails fast, does not degrade quietly

- no `torch_npu` / no usable device → `ImportError` / `RuntimeError` naming the
  cause, before the checkpoint is read
- libraries not built, or stale, or built for another part → refused at load
- `precision="fp8"` → `ValueError`; 910/A2 parts have no FP8 tensor hardware
- `precision` other than `"auto"` on another arch → `ValueError` naming what to
  use there instead
- unported options (`vision_pool_factor != 1`, `cache_frames != 1`,
  `state_prompt_mode="fixed"`) → `NotImplementedError`
- INT8 selected but not calibrated → `RuntimeError` saying what to call

## Receipts, on the part

Both tiers built through `load_model`, judged against an independent official
FP32 host. INT8 calibrated on 80 real frames across 40 tasks at percentile 90.
Two evaluation groups, calibration-disjoint by content hash: 16 in-distribution
and 40 held-out frames, one per task from separate episodes.

Per-frame full raw-action cosine:

| tier | group | minimum | median | below 0.995 |
|---|---|---|---|---|
| bf16 | in-distribution (16) | 0.999983 | 0.999997 | 0 |
| bf16 | held-out (40) | 0.999992 | 0.999997 | 0 |
| int8 | in-distribution (16) | 0.996836 | 0.999293 | 0 |
| int8 | held-out (40) | **0.994491** | 0.999231 | **1** |

The seven action channels track raw within 1e-4. All outputs finite.

**On the one frame.** BF16 clears the 0.995 bring-up target everywhere with three
orders of magnitude to spare and is the default tier. INT8 clears 0.99
everywhere, has a held-out median of 0.9992, and has one frame of 40 at 0.9945.
That frame's BF16 result is 0.99999, so the shortfall is the INT8 tier's own, not
this port's numerics. Percentile 99.9 and an earlier eight-frame calibration
recipe were both measured and neither removed it. INT8 is therefore opt-in and
documented on those terms; the tables are in `docs/deployment_npu.md` so the
trade is visible before anyone selects it.

Steady-state frame latency, production form — one prompt, one observation, graph
captured, timed after warm-up, 50 frames:

| tier | median | p90 |
|---|---|---|
| bf16 | 45.131 ms | 45.231 ms |
| int8 | **36.028 ms** | 36.176 ms |

Timings cover host upload, normalization, full model execution, output
unnormalization and download; they exclude checkpoint load, camera resize,
tokenization, calibration and capture.

**Determinism.** 300 replays of the captured INT8 frame give one distinct answer,
bit for bit, with the BF16 decoder arm through the same harness as the control.
Judged at 300 rather than a handful: at this site a broken arrangement passes a
five-replay check about half the time, and once did.

**Tests.** 88 passed, 1 skipped on the part. A public-path smoke confirms it
end to end there: `precision="fp8"` refused, `bf16` serves at cosine 0.9999963,
an uncalibrated `int8` `predict()` refused, `int8` serves at 0.9992928.

## Blast radius

59 files. Four of them are outside the NPU subtree, +91/−5 lines total:
`flash_rt/api.py` (the `precision` tier), `flash_rt/hardware/__init__.py`
(detection and the frontend registry), `flash_rt/core/precision_spec.py`,
`.gitignore`. No CMake, `.cu`, `.cuh`, HIP, AMD or RTX file is touched, and the
default NVIDIA and AMD kernel builds gain no NPU source.

One behaviour change is visible outside the NPU: `detect_arch` probes Ascend
before CUDA, because a CANN box reports `torch.cuda.is_available() == False`
while the part is perfectly usable. On a host carrying both, `hardware="auto"`
now selects the NPU; an explicit `hardware=` is never overridden. Documented in
the `detect_arch` docstring.

## Regression, on a machine with no NPU and no built CUDA extension

Same suite, same machine, base against this branch:

| | base | branch |
|---|---|---|
| passed | 1069 | 1120 |
| failed | **84** | **84** |
| skipped | 631 | 669 |
| collection errors | **8** | **8** |

The 84 failures and 8 collection errors are identical before and after; they come
from the unbuilt CUDA extension on that machine, not from this branch. The branch
only adds tests that pass or skip.

`tests/test_npu_contracts.py` is deliberately hardware-free — 33 tests a
reviewer without an Ascend device can run. They are routing tests rather than
signature checks: a fake frontend captures what `load_model` would construct, so
"int8 reaches the frontend", "bf16 turns it off" and "the part never receives
`use_fp8=True`" are asserted end to end. It also covers arch detection order,
the missing-runtime message, rejection of a stale or wrong-part shared object,
and that a device query which fails is a fault rather than a silently skipped
check.


LiangSu8899:
## Ascend 910B4 Pi0.5 — where the frame stands

One 910B4 die, two camera views, ten denoise steps. Every number is a **median
wall clock** in the served form — INT8 encoder, scales frozen from real
observations, captured graph — and every comparison is two arms in one process,
interleaved, with the parent measured twice so the gain has to clear its own
drift.

### Frame

| | ms |
|---|---|
| starting point of this branch's last two commits | 39.69 |
| after the encoder gate/up tile order | **38.46** |

The parent arm measured twice in the same session read 39.618 and 39.668, so the
1.18 ms clears its own spread twenty times over.

### What the last commits are worth

**`runtime(npu): walk the encoder gate/up tiles column block first`**

The kernel hands tiles to the twenty cube cores as `tile += 20`, so the twenty
indices in flight are consecutive. Walking the column index innermost put the
five tiles that share a column block thirty-two indices apart: no two tiles in
flight ever shared one, and every core streamed its own copy. The profiler reads
that off directly — `aic_mte2_ratio` 0.757 against `aic_mac_ratio` 0.413, load
bound with the cube 95% occupied waiting on it.

| | before | after |
|---|---|---|
| kernel median, 17 calls | 384.0 us | **296.4 us** |
| share of device time | 6503 us | **5027 us** |
| cosine vs the FP32 reference | 0.9992182 | 0.9992182 |

Each tile accumulates on its own, so the walk order cannot change the
arithmetic, and it does not — the cosine is identical to every digit.

**`runtime(npu): tile-align the SigLIP MLP and hold its weights in fractal NZ`**

SigLIP's MLP width is 4304, which is 16.81 fractal tiles of 256. Rounding to 17
whole tiles costs 48 zero channels and leaves the represented function unchanged;
both MLP weights move to fractal NZ at the same time. Paired over 16 real
observations: 41.47 → 39.44 ms median, **2.03 ms a frame, ahead on all 16**.
Accuracy against the FP32 reference is unmoved — worst of the 16 is 0.9966 here
against 0.9963 for the parent.

**`runtime(npu): read the encoder MLP down projection and SigLIP q/k/v from NZ`**

Fractal NZ pays on a wide K reduction and loses on a narrow one, so the layout is
chosen per site. The encoder MLP down projection reduces over 16384 columns and
moves; the 2048- and 256-wide projections stay ND. 39.67 → 39.47 ms median, ahead
on all 16 and **bit-identical to the parent on every one of them**.

### What just landed and is not yet called

**`runtime(npu): frozen INT8 decoder projections and their producer kernels`**

The four decoder projections run at ten action rows, where the whole cost is the
weight stream. This adds the kernels, their build stanza, the binding and its
tests; nothing in the serving path calls them yet.

The GEMM drives the cube directly — nd2nz into L1, one 2D load into L0, `Mmad`,
and the per-output-channel dequant applied by fixpipe on the L0C to GM path as
FP16. No tiling structure is loaded and the vector cores are untouched. Against a
reference built from the same frozen operands it sits at the FP16 output floor,
**cosine above 0.99999 at all four shapes**, in both the row-major and the
fractal-NZ output modes.

Three constraints hold it up and all three are silent when violated: 512-byte
write ownership between cube cores, an L0B slot that has to be exactly half the
buffer for a K split to accumulate, and a dequant vector staged once per launch.
They are described where they are enforced.

The GEMM declares itself cube-only at file scope, so it cannot share a
translation unit with the mixed gate/up kernel, and the kernel headers define a
per-unit tiling symbol, so it cannot share a library either — it gets its own
build stanza. The three vector siblings join the dispatch unit, and the shipped
kernels they now sit beside are unaffected: the gated AdaRMS next door measures
**2.44 us before and 2.44 us after**, to the hundredth, over its 370 calls.

Suite on the device: **44 passed, 1 skipped**.

### Next

Wiring those projections into the serving path takes the frame to **36.7 ms**
measured in place, at the parent's accuracy. That wiring is not in this branch:
it still has an open question about the first replay after graph capture
differing from every replay after it, and it is not going in until that is
settled.

LiangSu8899:
## The decoder projections are in, and a kernel bug came out with them

Same die, same two views, same ten denoise steps, same served form as above.

| | ms | replays disagreeing |
|---|---|---|
| BF16 decoder | 38.68 | 0/300 |
| **INT8 projections** | **36.88** | **0/300** |

Cosine against the FP32 reference is **0.9992910**, against 0.9992182 for the
BF16 decoder it replaces — closer to the reference, not further. Together with
the encoder tile order earlier in this branch the frame is **39.69 → 36.88**.

### The kernel bug

The INT8 GEMM writes two of the four projections in fractal NZ, and it was
writing only the ten live rows of each sixteen-row block. A block interleaves its
rows at 32-byte granularity, so writing ten of them threads gaps through a
1024-byte block, and between cores those partial writes do not settle. With every
projection running, a captured frame stops replaying to a single answer.

Three hundred replays on identical input, one variable:

| | replays disagreeing | distinct answers | ms |
|---|---|---|---|
| live rows only | 38/300 | 19 | 36.879 |
| **whole block** | **0/300** | **1** | **36.819** |

Writing the whole block costs nothing measurable and the rows past the live ones
are never gathered — the consumer walks rows zero to M-1.

### On the gate that found it

The defect was there from the first version of these kernels and the check that
signed them off did not have the power to see it. A replay disagrees about one
time in eight, so a check comparing five replays passes about half the time
whether the arrangement is sound or not; sixty replays cannot separate a fixed
arrangement from a few per cent left over.

Three hundred replays cost eleven seconds against the four minutes of setup
around them. Everything above is measured at three hundred, with the BF16
decoder through the same harness as the control, and it is exact.

Two smaller things fall out. Every layer shares one result buffer per projection
rather than holding its own, which is what puts a write-after-read between
adjacent layers. And the decoder step reads each AdaRMS branch's layout off the
projection that produced it rather than assuming one, so the two cannot drift
apart.

### Calibration

The decoder runs each projection once per denoise step and the activation
magnitude moves along the ODE trajectory, so one scale for the whole trajectory
spends most of its range on steps that do not need it. The observer is per step —
the step index is the call index, since the decoder visits every layer once per
step in step order — and the pass is shaped like the encoder's: one eager sample
at a time, observers reset between samples, every site has to have run. It runs
after the encoder is bound, so the maxima it sees are the ones the served frame
will see.

Suite on the device: **45 passed, 1 skipped**.

LiangSu8899:
### Decode attention over a transposed value cache — 36.91 → 36.16 ms

`0a18503` moves the action decoder's attention off the vendor operator and onto
the raw cube path, over a KV cache that keeps the values transposed in fractal
NZ. A raw `Mmad` B operand wants its GM source in `(N, K)` form, which for
`O = P * V` means V as `(HD, KV)`; keeping the transpose in the cache makes both
attention GEMMs the same code path as the decoder GEMM, with nothing to convert
on either operand.

Both arms are the production form through the same harness, differing only in
the runtime switch:

| | vendor operator, row-major cache | transposed cache |
|---|---|---|
| wall p50 | 36.911 ms | **36.159 ms** |
| wall min | 36.845 ms | 36.116 ms |
| device total | 34886 us | 34056 us |
| attention | 25.63 us × 180 = 4634 us | **19.16 × 180 = 3446 us** |
| rotary | 4.12 × 180 = 754 us | 6.04 × 180 = 1063 us |
| prefix transpose | — | 3.85 × 18 = 78 us |
| cosine vs FP32 reference | 0.9992910 | **0.9992928** |

The layout costs no launches: the value half of the once-a-frame prefix fill
becomes the transposing kernel rather than gaining one, and the rotary that
already wrote the value rows writes them transposed.

Two details are load-bearing rather than incidental. The transpose is stored in
fractal NZ so that a sixteen-position block is 8 KB of contiguous bytes — that
is what makes a denoise step's ten new value columns one aligned write instead
of 256 scattered two-byte ones, and it also removes the `Nd2Nz` conversion from
the operand that dominates the kernel's reads. And the action suffix now leads
the cache with the encoder prefix starting at column 16: the prefix is not a
multiple of sixteen rows, so a suffix placed after it would straddle two fractal
blocks and have cores writing parts of blocks their neighbours also hold.
Attention does not care in what order the keys arrive.

**Determinism:** 300 replays of the captured frame agree to the bit, one
distinct answer, with the vendor arm as the control through the same harness.
**Tests:** 55 passed, 1 skipped.

Also measured and not taken: splitting the attention over the key range. It is
arithmetically exact at one, two and four splits and takes `cube_utilization`
from 24% to 94.6%, and it is not faster. Sweeping the key length shows the cost
is affine in it — 9.44 us fixed plus 0.0101 us per KB across seven points from
64 KB to 2048 KB — so at this geometry 9.4 of the kernel's 19.2 us is spent
before it does any work. A split compresses only the remainder and pays a second
launch to do it.

PR #200, opened 2026-09-12
npu: serve GR00T N1.7 on Ascend 910B4 (396.67 → 52.40 ms, cosine 0.9999570)

LiangSu8899:
Adds a native Ascend 910B4 path for GR00T N1.7 (Qwen3-VL ViT + truncated
Cosmos-Reason2 LLM + VL adapter + 32-layer DiT action head), routed as
`("groot_n17", "torch", "npu")`. The whole model is one captured graph with three
inputs — camera patches, robot state, noise draw — plus three native kernels: the
DiT's attention, a fused add-and-normalise, and the evaluation image transform's
resize.

## Results

Full frame, end to end: one observation in, denormalized robot-space actions out.
Ascend 910B4, CANN 8.5.2, PyTorch 2.7.1 + torch_npu 2.7.1.post2.

| | |
|---|---|
| PyTorch eager, official policy | 396.67 ms |
| **this path** | **52.40 ms** (p90 52.61) — **7.6×** |
| combined denormalized-action cosine | **0.9999570** |
| device memory | 6.09 GB |

`torch.compile` was not measured on this backend, so the judged baseline is eager.
Per modality: end-effector 0.9998915, joint 0.9999829, gripper 0.9956549 (nearly
constant over a trajectory, not a useful gate). 300 replays of the captured chain
give one distinct answer, bit-identical to the same graph run eagerly.

## Build

Default is unchanged — the four shared objects Pi0.5 loads, ABI still 1. This
model's units are opt-in:

```bash
FLASHRT_ENABLE_NPU_GROOT_N17=ON bash scripts/npu/build.sh
```

## Documentation

- `docs/deployment_npu_groot_n17.md` — usage, the prompt bundle's keys, what each
  kernel does and why, precision, tests
- `docs/deployment_npu.md` — the shared Ascend build
- `csrc/npu/README.md` — the three entry points, their envelopes and refusals

## Tests

`python -m pytest tests/test_npu_*.py -q` → 146 passed, 52 skipped without a
device. `tests/test_npu_groot_n17_device.py` holds the kernels' numerics and needs
the part.

---

Draft: the figures above were measured before this branch's last revisions
(shared-object names, operand checks, build selection, the frontend) and are being
re-taken on this head.


LiangSu8899:
Thanks — this was a useful review. Five of the six blockers are fixed; the sixth
is real and I have converted the PR to a draft rather than argue it. One naming
point I am pushing back on, with evidence. Head is now `7034dcc`.

## 1. Model-level compile isolation — fixed

`FLASHRT_ENABLE_NPU_PI05` (default `ON`) and `FLASHRT_ENABLE_NPU_GROOT_N17`
(default `OFF`), exactly as suggested. The default build emits the same four
shared objects it did before these units existed; a selection with neither model
is refused rather than silently producing nothing; a value that is not `ON`/`OFF`
is refused by name.

The split is clean because GR00T loads **none** of the four existing libraries —
it uses `NpuGraph`, `abi` and `NzBf16Weight`, none of which open a shared object:

```
pi05=ON  groot=OFF -> libflashrt_npu.so libflashrt_npu_cube.so libflashrt_npu_decoder.so libflashrt_npu_attn.so
pi05=OFF groot=ON  -> libflashrt_npu_dit_norm.so libflashrt_npu_image.so libflashrt_npu_dit_attn.so
```

## 2. ABI — fixed, back to 1

You are right and the reason is worth writing down, so I put it next to the
constant: the number tracks what has **shipped**, not how many times an
unreleased signature was edited in a branch. Relative to `main` this branch
changes no already-published entry point. Reverted to 1 in `abi.py` and in the
build script's default.

## 3. Raw pointers and device residency — fixed, and it found a second bug

New `flash_rt/npu/core/operands.py` holds the checks instead of each call site
restating them. They read tensor attributes only, so they are tested on a host
with no Ascend device (`tests/test_npu_operands.py`).

`require_npu` exists for the case you are pointing at specifically: "same device
as that other operand" passes a launch whose operands are *consistently* on the
host and still faults, so residency is checked separately and returns the device
the rest are held to.

- the image transform requires its frames on the device and says to copy them
  there; the doc's example now uploads explicitly, and the upload is called out
  as part of the frame's cost
- the norm checks residual, branch, the affine pair, the branch bias and the
  padded destination against each other, and `serves()` answers `False` for a
  host tensor so routing cannot reach the kernel with one
- the attention's checks were already there and now go through the same helper,
  so all three name the operand that was wrong
- a CPU tensor is rejected by test, in both wrappers

**And the check you asked for surfaced something worse.** The transform only
refused the case where a resize did *nothing*. It reproduces OpenCV's
**enlarging** `INTER_AREA`; shrinking is a different branch inside OpenCV (true
area interpolation over a variable number of taps). So a camera whose smallest
edge already exceeds 256 — a 480×640 sensor, say — was resized with the wrong
weights and **no error**. Both steps must now enlarge, and the tap builder
refuses a shrinking pair itself. There is a test for each.

## 4. Not a complete model integration — conceded, PR is now a draft

I have taken your option 2 for the document and stopped short of claiming the
rest. The page now opens by calling itself compute groundwork rather than a
deployment, names the one missing piece, and gives the argument for finishing it
instead of landing it: until the frontend exists, every deployment writes that
assembly again. The example no longer has undefined names — it says where
`processor`, `embodiment_id`, `action_dim`, `height/width/views` come from,
splits setup from the per-frame path, and tabulates the seven keys `aux` has to
carry.

I am **not** asking you to accept un-routed code into main, so the PR is a draft
until the frontend and its `_PIPELINE_MAP` entry land. One thing I want to be
straight about rather than let it look like a choice: the 910B4 this was measured
on is currently unreachable, and that frontend has to reproduce the official
processor's observation bundle — the prompt's token partition, the two rope
tables, the interpolated patch positions, the instruction's embeddings. Writing
roughly a thousand lines of that blind, with no way to run it end to end, is the
one thing this work has been punished for before. It is the next change, with an
on-device smoke through `load_model`, not this one.

## 5. Dependency direction — fixed

`flash_rt/models/groot_n17/weight_spec.py` now holds the spec, beside that
model's other backend-independent pieces. It was already shared — the RTX, SM89
and FP16 frontends all imported it too, under a name that said Thor — so the
move is where it should have been from the start.
`flash_rt/frontends/torch/_groot_n17_thor_spec.py` stays as a re-export, so every
existing importer is untouched and nothing about Thor's behaviour moves; the test
asserts the two names are the same object.

## 6. Tests — fixed, 51 added

`114 passed, 54 skipped` on a host with no Ascend device, against the 63/39 you
measured.

| file | needs a device | covers |
|---|---|---|
| `test_npu_operands.py` | no | host tensor, wrong device index, dtype, free-axis shape, row pitch, transposed operand, non-tensor |
| `test_npu_groot_n17_units.py` | no | each of the three loaders: binds its entry point on a matching identity, refuses a stale ABI, refuses another part, says to run `build.sh` when unbuilt. The norm's refusals. **The tap construction against OpenCV itself**, four enlarging geometries, bit for bit |
| `test_npu_groot_n17_actions.py` | no | the action decode and state encode with a fake processor and `device="cpu"`: the absolute inverse, the clamp, relative joints, a relative end-effector pose composed onto its reference frame, modality order, and every representation they refuse |
| `test_npu_groot_n17_device.py` | yes | attention against the vendor operator at all three DiT geometries and its two value layouts against each other bit-for-bit; the norm against a reference that rounds where it rounds, its folded bias against the separate add, its padded write; the whole image transform against OpenCV |

The image transform's exactness is now checked **without** a device, which is the
part that was wrong three times: it is host arithmetic, so OpenCV can be the
oracle on any machine.

**Disclosed:** `test_npu_groot_n17_device.py` is written but has **not been
executed** — the part is unreachable, so it currently only collects and skips.
That is the first thing to run when it is back, and I would rather you knew than
find a permanently-skipping file later. Happy to drop it from this PR if you
would rather it landed only once green.

## Naming — half taken, half argued

`dit_vector` was fair: it named a *core type*, not an operation. The unit is now
`dit_norm` — `dit_norm_910b.cpp`, `libflashrt_npu_dit_norm.so`,
`FLASHRT_NPU_DIT_NORM_LIBRARY` — matching the entry point, which was already
`flashrt_npu_dit_add_layer_norm`.

`dit_attn` I would like to keep, because renaming it to `groot_n17_dit_attn`
would be the odd one out in this subtree and would contradict the convention it
follows. Its neighbours are all Pi0.5-only and all named for the operation:

```
csrc/npu/kernels/gu_int8_910b.cpp        gate/up INT8      (Pi0.5 only)
csrc/npu/kernels/decoder_gemm_910b.cpp   decoder GEMM      (Pi0.5 only)
csrc/npu/kernels/decode_attn_910b.cpp    decode attention  (Pi0.5 only)
```

"DiT" is an architecture, not a model, the same way "decode" is a phase. What I
think you were actually pointing at is the ambiguity — is this a general
primitive or not — so I have resolved that where it is checkable rather than in
the filename: the wrappers stay under `flash_rt/npu/models/groot_n17/` because
only that model's geometries are validated, every kernel refuses anything outside
its envelope with a code, and `csrc/npu/README.md` now documents each entry
point's envelope and its refusals. If you would still rather have the model in
the filename, say so and I will move all three.

## Docs synced

- `abi.py` and `csrc/npu/abi.h` no longer say "four shared objects"
- `csrc/npu/README.md` documents the three units' entry points, envelopes and
  refusals, plus the build switches
- `docs/deployment_npu.md`'s build section is a table of the two switches; it had
  said seven by default, which the switches make wrong

## Local

```text
no-device tests:  114 passed, 54 skipped
compileall:       PASS
diff --check:     PASS
build.sh -n:      PASS  (and the ON/OFF guards exercised)
```

Unchanged: no CUDA, ROCm or root CMake target is touched, and no shared frontend
or pipeline is forked.

LiangSu8899:
Head is now `4bf088f`. Points 2, 3 and 5 are fixed; 1 and 4 I agree with and the
PR stays a Draft because of them.

## 2. Model attribution for the kernels — taken, and you were right

My last answer gave the convention this subtree follows and did not answer the
argument you actually made: a second DiT or WAM path has an equal claim on
`dit_attn`, and at that point a file with no model attribution is ambiguous rather
than merely terse. "The old files are named this way" is not a reason. Withdrawn.

I took the directory option rather than the filename prefix, because it gets both
properties at once:

```
csrc/npu/kernels/groot_n17/dit_attn_910b.cpp     -> libflashrt_npu_groot_n17_dit_attn.so
csrc/npu/kernels/groot_n17/dit_norm_910b.cpp     -> libflashrt_npu_groot_n17_dit_norm.so
csrc/npu/kernels/groot_n17/area_resize_910b.cpp  -> libflashrt_npu_groot_n17_image.so
```

The model owns the directory **and the artifacts**, so the ambiguity you named
cannot arise — the next DiT path gets its own directory rather than competing for
the name — while the file names stay operation names like everything else here.
Overrides became `FLASHRT_NPU_GROOT_N17_DIT_ATTENTION_LIBRARY`,
`..._DIT_NORM_LIBRARY` and `..._IMAGE_LIBRARY`. None of these artifacts have
shipped, so the rename costs no user anything.

If you would still rather have the model in the file names as well, say so and
I will add it.

## 3. Stale artifacts — fixed, and you described the failure exactly

A selection that only decides what to *add* is not a selection. Build with GR00T
on, build again with the default, and the directory held seven libraries while the
script and the docs said four — and a loader would have bound them.

A deselected model's standard artifacts are now removed and each removal is
reported. Only the names this script produces are touched; an override lives
wherever the caller put it, and an unrelated file in the directory is left alone.

Your "补顺序构建测试" is `tests/test_npu_build_selection.py`, eight cases against
the real script, needing neither CANN nor a device — `FLASHRT_NPU_PRUNE_ONLY=1`
performs the selection and reports without invoking a compiler:

| case | asserts |
|---|---|
| GR00T on, then the default | exactly the four Pi0.5 libraries remain, and each removal is reported rather than silent |
| the reverse selection | exactly the three GR00T libraries remain |
| both on | nothing is removed |
| an unrelated file and a custom `.so` present | untouched |
| neither model selected | refused **and nothing deleted** |
| a switch that is not `ON`/`OFF` | refused by name |
| — | the selection appears in the output |
| a part these kernels are not validated for | refused |

Two smaller things that fell out of writing it. A switch set to the **empty
string** was silently taken as the default (`${VAR:-ON}`); it is now refused like
any other bad value. And a refused selection must not have deleted anything on the
way to refusing — there is a test for that, because the prune runs before the
build.

## 5. PR description — rewritten

You were right that it was describing the first version. It now opens with the
Draft reason, has the current build table, the current shared-object names, says
the ABI stays at 1, lists the six test files with which need a device, and has a
**"Still to do on hardware"** checklist that is exactly your point 4.

One thing I added rather than leave implied: the 52.40 ms / 0.9999570 figures were
measured **before** the last two review rounds, which changed shared-object names,
added argument checks in front of every kernel and changed what the build emits.
They have to be re-measured on the current head before being quoted as final, and
the description now says so where the numbers are.

## 1 and 4 — agreed, which is why it stays a Draft

**1.** Nothing to argue: without the frontend and its `_PIPELINE_MAP` entry this
does not meet the new-model bar, and it should not be on main. The document says
so at the top and the description now says so above the fold.

**4.** Your list is the right list, and it is in the description as a checklist
because this round changed shared-object names, the ctypes argument checks and the
build selection — so the device file has to be run on **this** head, not on the
one the numbers came from. `tests/test_npu_groot_n17_device.py` is still
**written and never executed**; the 910B4 has been unreachable throughout both
review rounds. Standing offer: if you would rather not carry a
permanently-skipping file, I will drop it and land it green with the frontend.

## Local, on this head

```text
tests/test_npu_*.py:  123 passed, 52 skipped     (was 112 / 52)
with weight_loader:   125 passed, 54 skipped
compileall:           PASS
bash -n build.sh:     PASS
git diff --check:     PASS
```

Unchanged from the last round: the default build emits the same four shared
objects as before this branch, the ABI is 1, no existing Pi0.5 Python or runtime
path is rewritten, and no CUDA/ROCm/root-CMake target is touched.

LiangSu8899:
Squashed to **one commit**, `efacaf8`. Blocker 1 and the `PRUNE_ONLY` note are
fixed; blocker 2 is the only thing left and it needs the part, which is still
unreachable — so this stays a Draft and I am not claiming otherwise.

## 1. Frontend and routing — done

`("groot_n17", "torch", "npu")` → `GrootN17TorchFrontendNpu`, and
`docs/stable_api.md` now lists `npu` (and `amd_cdna4`, which was also missing from
that enumeration).

The contract is the one this model already has on its other backends, which turned
out to matter: `VLAModel.predict` calls `set_prompt` with a prompt string alone,
and every GR00T N1.7 frontend needs the observation bundle as well — so Thor and
CDNA4 are both driven through `model.pipeline`, and the bundle is the caller's on
both. **I had that wrong in an earlier reply**, where I described the missing piece
as "a frontend that reproduces the observation bundle". It does not; no backend's
does. That made this far smaller than I estimated, and it is the reason the
frontend is 253 lines rather than the thousand I quoted.

```python
fe = model.pipeline
fe.set_hf_processor(processor)
fe.set_prompt(aux=aux, prompt=instruction)
state = fe.normalize_state(state_dict)
out = fe.infer(state, aux={"frames": frames})
actions = fe.denormalize_action(out, state_dict=state_dict)
```

Two things it refuses rather than approximates. `precision="int8"` raises instead
of quietly serving BF16 — a downgrade nobody was told about is measured as a
regression. And `infer` without a fresh observation raises, because this tier
captures the whole frame in one graph and a features-only replay is a second graph
that has not been measured; I would rather it said so than returned the previous
frame's answer.

`tests/test_npu_groot_n17_frontend.py` — 21 cases, no device: the routing, the
contract's shape against the other backends', and every argument refused (an
unknown embodiment, a view count that embodiment was not trained with, a prompt
bundle missing a key *named in the error*, a bundle describing other cameras,
`infer` before `set_prompt`, both or neither of `frames`/`patches`, a host frame
tensor, the noise draw's shape, and both host-side halves before the processor is
set).

## `FLASHRT_NPU_PRUNE_ONLY` — removed, and you were right to call it a backdoor

Gone from the script entirely. The tests now drive the **real** script end to end
against a fake toolchain — a `bisheng` and a `c++` that create whatever their `-o`
names and compile nothing. That is better than the hook in two ways: the script
carries no test-only path, and the tests now also check the things the hook
skipped, namely that the sources come from `csrc/npu/kernels/groot_n17/` and that
the artifact names are the ones the loaders look for.

13 cases, including both sequential-build directions, the refused selection that
must not have deleted anything on its way to refusing, and an override and an
unrelated file left alone.

## 2. Receipts on this head — not done, and I cannot fake it

The 910B4 has been unreachable throughout these review rounds (port closed, three
attempts this round). Everything in your list is still outstanding and is a
checklist in the PR description, with one addition of my own: the full-frame smoke
now has to go **through `load_model`**, because that exercises the frontend for the
first time.

- [ ] default Pi0.5-only build, and the existing Pi0.5 suite
- [ ] `FLASHRT_ENABLE_NPU_GROOT_N17=ON` build of the three units
- [ ] `tests/test_npu_groot_n17_device.py` — written, never executed
- [ ] one full-frame parity and capture smoke through `load_model`

The description also marks the 52.40 ms / 0.9999570 figures as measured on the
compute *before* these rounds, since they changed shared-object names, the
argument checks, the build selection, and now added the frontend.

## Squashed

One commit, 31 files, +5494 / −468. The intermediate history was review churn and
is gone.

```text
tests/test_npu_*.py:  146 passed, 52 skipped     (123 / 52 last round)
compileall:           PASS
bash -n build.sh:     PASS
git diff --check:     PASS
```

Unchanged: the default build emits the same four shared objects as before this
branch, the ABI is 1, no existing Pi0.5 Python or runtime path is rewritten, and
no CUDA/ROCm/root-CMake target is touched.

PR #201, opened 2026-09-13
pi05: opt-in noise export and denoise tracing

LiangSu8899:
## Scope

This PR now contains only seeded noise injection and opt-in denoise tracing for Pi0.5 RTX. Tokenizer normalization and batch/prefix features have been removed from this PR and moved to #203/#204.

- `return_noise=False` by default: ordinary inference never downloads the noise tensor.
- Explicit `return_noise=True` returns the actual quantized initial noise used by the model.
- `noise=` and trace collection are explicit; ordinary stochastic inference keeps the existing path.
- Batch staging and noise updates use the graph's execution stream.
- No new CUDA source or public kernel symbols.

## Validation

RTX 5090 container, Python 3.12, PyTorch 2.9.0a0+145a3a7bda.nv25.10, CUDA 13.0, Pi0.5 LIBERO checkpoint:

`pytest -q tests/test_pi05_seeded_noise_trace.py tests/test_pi05_noise_pooling_contract.py`: **4 passed, 0 skipped**, on this intermediate branch with the baseline extension.

Tests cover reproducible injection/export, trace consistency, trace disabled by default, batched per-slot behavior, and absence of default noise host copies. Python compilation and `git diff --check` pass.

The later #202 branch, which includes these changes with its optimization disabled, produces bit-identical final actions to main on the 400-call real-observation E2E comparison. Full container timing and the stack's numerical caveats are recorded in #202.

## Review order

The original two large PRs are now split into focused stacked PRs:

1. #201: seeded noise and opt-in denoise trace.
2. #203: production tokenizer normalization.
3. #204: batch widths and valid-length prefix export.
4. #202: opt-in skinny decoder.
5. #205: safe weight reload.
6. #206: opt-in NVFP4 prefix.
7. #207: prompt and batch-width context caches.
8. #208: opt-in SDE sampler.

Each PR targets its immediate predecessor so its Files changed view contains only its feature. Merge in this order and retarget descendants after the parent lands; squash-merging a parent requires rebasing the next PR first. No PR has been merged automatically.

PR #203, opened 2026-09-17
pi05: normalize SentencePiece prompts through a tested production helper

LiangSu8899:
## Scope

Split out of #201. Depends on #201; this PR changes only the SentencePiece prompt-normalization contract.

- Share the production `encode_pi05_prompt` implementation with the frontend fallback.
- Match the intended lowercase/whitespace/underscore and BOS/newline formatting.
- Tests call the production helper, not a duplicate normalization implementation.
- No kernels, compilation changes, batching, or sampling changes.

## Validation

RTX 5090 container, Python 3.12, PyTorch 2.9.0a0+145a3a7bda.nv25.10, CUDA 13.0:

`pytest -q tests/test_pi05_prompt_tokenization.py`: **5 passed**.

Python compilation (excluding the repository's intentionally incomplete frontend template) and `git diff --check` pass.

PR #204, opened 2026-09-17
pi05: batch-width support and valid-length prefix feature export

LiangSu8899:
## Scope

Split out of #201. Base: `feat/pi05-tokenizer-normalization`. Adds arbitrary positive batch widths and opt-in prefix feature export only.

- Retain per-slot prompt lengths; pooling excludes each slot's padding.
- Preserve replay-stream ordering for image/noise staging.
- Add unequal-length prompt prefix comparisons against independent single-sample inference.
- No default feature export or noise download. No new CUDA translation units or symbols.

## Validation

`pytest -q tests/test_pi05_batched_n.py tests/test_pi05_noise_pooling_contract.py`: **6 passed, 0 skipped**.

The independent baseline-kernel build is used for this branch. Python compilation and `git diff --check` pass.

Known legacy limitation: the asymmetric CFG slot-1 cosine gate also fails on unchanged main in the earlier host regression (0.980458 vs 0.985; audited sampling branch 0.982395). No threshold was relaxed. This PR does not claim to fix that existing CFG precision issue.


LiangSu8899:
## Follow-up verification and merge record

Rebased onto main after #201 and #203 were squash-merged, then retargeted to main.

- Corrected B=2-only documentation in the frontend, batched pipeline and attention backend. `_b2` is retained as a historical internal name; general batching supports N >= 1, while CFG still requires exactly two slots.
- Added B=4 mixed observations and independently seeded per-slot noise, compared with four independent native B=1 calls. Calibration input, seed and activation-scale arrays are checked to match.
- Added a graph-reuse permutation check: moving every observation/noise pair to a different slot preserves each final action bit-for-bit.
- Added the missing input-stream completion barrier before the batched calibration routine enters its stream-0 parent forward. This is setup-only, not a new hot-path synchronization.

RTX 5090 / CUDA 13.0 container, PyTorch 2.9.0a0+145a3a7bda.nv25.10, Pi0.5 LIBERO checkpoint:

`pytest -q tests/test_pi05_batched_n.py tests/test_pi05_noise_pooling_contract.py tests/test_pi05_batched_precision.py`: **8 passed, 0 skipped**.

Numerical contract clarification: the initial new test used a provisional 0.999 cosine gate. The established synthetic-observation gate in `test_pi05_batched_precision.py` is 0.99; the new test follows that existing contract without changing any prior test threshold. With identical calibration, the mixed FP8 slots measure 0.99988794, 0.99974081, 0.99840498, 0.99371511 against native B=1. Existing B=2 on those same inputs measures a minimum of 0.99403108; B=4 versus B=2 has minimum 0.99968335. Slot permutation is exact. Documentation no longer promises unconditional 0.9999 B=N/B=1 agreement.

An additional BF16 diagnostic on these white-noise images is not uniformly close to native B=1 (minimum about 0.882 for existing B=2 and 0.900 for B=4; B=4/B=2 minimum 0.99916). This is not a claim of general BF16 numerical equivalence or robot task success; the new accepted mixed-input numerical gate is the existing FP8 contract.

Diff whitespace checks pass. The PR was merged only after its focused tests passed.

PR #202, opened 2026-09-13
pi05: opt-in SM120 skinny decoder with isolated kernels

LiangSu8899:
## Scope

This PR has been reduced to the Pi0.5 skinny FP8 decoder family (GEMMs, fused consumers, attention and step projections). It depends on #204. Reload, NVFP4 prefix, prompt/context caching and SDE are separate PRs, not part of this diff.

- `FLASHRT_ENABLE_PI05_SKINNY=OFF` by default; enabling requires `GPU_ARCH=120`.
- Source, header, object target and pybind symbols share the Pi0.5 feature gate, independently of Qwen XQA.
- Model-owned bindings use `pi05_dec_skinny_*` names.
- Both frontend and pipeline retain `decoder_kernel="cublaslt"` by default. Default construction does not create skinny transposed weight copies.
- The alternative requires explicit `decoder_kernel="skinny"` (or an explicit environment override).
- The benchmark measures synchronized complete observation-to-final-action wall time; obsolete graph/kernel performance tables were removed.

## Validation

RTX 5090 container, Python 3.12, PyTorch 2.9.0a0+145a3a7bda.nv25.10, CUDA 13.0, Pi0.5 LIBERO checkpoint:

Independent skinny-ON SM120 build/link/import passed.

`pytest -q tests/test_pi05_decoder_skinny.py tests/test_pi05_skinny_build_gate.py`: **12 passed, 0 skipped** on this intermediate branch. Covers default routing/allocation, fused-consumer rounding, GEMM/attention/projection parity, final actions and deterministic graph replay.

Python compilation (excluding the pre-existing intentionally incomplete frontend template) and `git diff --check` pass.

The final split stack passes **49 tests, 0 skipped** across 13 Pi0.5 files in this container. Its core was rebuilt and imported with all Pi0.5 feature options OFF for **SM87/89/110/120/121**; generated source lists and bindings contain none of the new Pi0.5 kernels. All-ON SM120 and skinny-ON/SDE-OFF SM120 build/import/symbol checks also pass. These are cross-build/import checks for non-SM120 targets, not execution on those GPUs. The pre-existing FA2 module imports with every core; the matrix does not claim to rebuild FA2 for each architecture.

## Synchronized E2E

[...]

PR #205, opened 2026-09-17
pi05: serialized weight reload with preflight and failure invalidation

LiangSu8899:
## Scope

Split out of #202. Depends on #202. This PR adds serialized, preflighted Pi0.5 weight reload; no new kernel sources or build options.

- Validate converted key coverage, shapes, dtypes and supported modes before touching live buffers.
- Serialize reload, inference, prompt updates and calibration with an instance-local reentrant lock.
- Keep the instance usable after preflight failure. If mutation has begun and any later step fails, mark the frontend unusable and require reconstruction; never continue inference with mixed weights.
- Keep graph-owned addresses and refresh FP8 scales, decoder styles and prompt embeddings.
- Do not call process-wide `torch.cuda.empty_cache()`; allocator policy belongs to the caller.

## Validation

`pytest -q tests/test_pi05_weight_reload.py tests/test_pi05_reload_lifecycle.py`: **6 passed, 0 skipped**.

Includes malformed/partial checkpoints, unchanged-instance behavior on preflight failure, reload-versus-fresh output comparisons, batched buffers, mutation-failure invalidation, and mutual exclusion. Python compilation and `git diff --check` pass.


LiangSu8899:
Rebased onto main after #201 -> #203 -> #204 -> #202 were squash-merged, and retargeted this PR to main. Dependent #206-#208 branches were also rebased and pushed; none has been merged.

Please keep the downstream stack on hold: final all-enabled integration testing produced 50 passed / 1 failed, with a reproducible B=4 versus B=1 mixed-slot numerical mismatch. The merged-main control passed all 15 tests in the matching prefix sequence. The failure is not yet attributed to a specific downstream PR. Full evidence: https://github.com/flashrt-project/FlashRT/pull/208#issuecomment-5711506005

LiangSu8899:
## Mixed-slot numerical mismatch: root cause and fix

The failure existed in the original, unsplit #202. It was not introduced by splitting or rebasing the PR stack.

### Controlled reproduction

RTX 5090, CUDA 13 container, identical checkpoint, observations, initial noise and calibration seed. The historical Python trees were tested against the same compatible native extension with the decoder explicitly pinned to cuBLASLt. This isolates the Python regression; it is not a claim that every historical native build/profile was rebuilt or tested.

| Tree | Minimum B=4 versus independent B=1 action cosine |
| --- | --- |
| Original #201, `a7099689` | 0.9937151057 |
| Original #202, `dcc23522` | 0.0748627863 |
| Split #205 before fix, `52d3094a` | 0.0748627863 |
| Split #205 after fix, `610450eb` | 0.9937151057 |

In each case the four slots have different synthetic observations and initial noise. Calibration scales match between B=1 and B=4, and permuting the slots preserves their outputs bit-for-bit. After the fix, **both the four B=1 outputs and the four B=4 outputs are bit-identical to the original #201 outputs**.

### Root cause

Original commit `56a996e7` replaced the FP8 weight quantizer's Python-double scale calculation and scalar division with FP32 device scale arithmetic and tensor division. That changes rounding at FP8 midpoint boundaries. Converted BF16 checkpoint tensors themselves are equal; the quantization change is sufficient to cause the observed downstream divergence.

The fix retains device-only execution: compute the scalar scale in FP64, cast to FP32, and multiply by its reciprocal to preserve the original scalar-division contract. There is no `.item()`, CPU transfer, new inference synchronization, kernel change, or relaxed numerical threshold.

### Verification

- Original #201 related tests: **11 passed**.
- Original #202 related tests on the controlled cuBLASLt route: **19 passed**. Those older tests do not exercise the newly added mixed-slot regression, which independently reproduces the failure above.
- Added **9 tests** for exact legacy quantized bytes/scales, zero and small scales, noncontiguous input, and CUDA Graph capture/replay with changed input.
- Fixed #205 quantization plus mixed-slot tests: **10 passed**.
- Full updated #205-#208 stack, including seeded noise, pooling, tokenization, mixed batching, skinny kernels, build contracts, reload lifecycle, NVFP4, prompt/width caches, SDE and the new quantizer tests: **60 passed, 0 skipped**.
- Changed Python compilation and `git diff --check`: passed.

The correction has been propagated through #205-#208. This supersedes the earlier unresolved 50-pass/1-fail receipt. No additional PR has been merged. These results establish compatibility for the tested inputs, not universal BF16/FP8 equivalence or an official-model accuracy claim.

LiangSu8899:
Merged after verifying the audited reload/FP8 compatibility fix at 610450eb. The preceding CUDA 13 / RTX 5090 regression receipt is 60 passed, 0 skipped. Following the requested stack order, #206 is now rebased onto the new main and retargeted to main; the other PRs will be processed sequentially.

PR #206, opened 2026-09-17
pi05: opt-in SM120 NVFP4 prefix tier with model-owned build gate

LiangSu8899:
## Scope

Split out of #202. Base: `feat/pi05-weight-reload-safe`. Adds only the opt-in Pi0.5 NVFP4 vision/prefix tier.

- `FLASHRT_ENABLE_PI05_NVFP4=OFF` by default; enabling requires `GPU_ARCH=120`.
- Source, header, compile definition and `pi05_geglu_merged_to_nvfp4_swizzled` binding share the same model-owned gate.
- Runtime remains FP8 by default. Explicit `prefix_precision="nvfp4"` fails clearly if its fused kernel was not built.
- Weight reload refreshes the selected NVFP4 representation in place.
- Add an explicit NVFP4 arm to the synchronized observation-to-action benchmark. No partial graph/kernel timing is presented as E2E.

## Validation

Independent SM120 build/link/import with skinny + NVFP4 ON succeeds, without SDE sources or bindings.

`pytest -q tests/test_pi05_prefix_nvfp4.py`: **3 passed, 0 skipped**. Covers fused quantizer parity, FP8/NVFP4 final actions, deterministic replay, batched inference and reload.

Python compilation and `git diff --check` pass. NVFP4 is approximate and requires task-specific qualification; it is not made the default.


LiangSu8899:
Review follow-up complete: the Pi0.5-only GEGLU NVFP4 source/header now live under csrc/kernels/pi05/ with pi05_ filenames; CMake and binding includes match. The NVFP4 option now prints explicit ENABLED/DISABLED status. After #205 merged, this branch was rebased onto main and retargeted to main; its diff contains only the NVFP4 increment.

RTX 5090 / CUDA 13: rebuilt the complete all-enabled core extension successfully. The final stacked regression suite passed 66 tests with no skips, including NVFP4, reload, mixed-slot batching, skinny, caches, SDE and FP8 byte compatibility. All-OFF configuration excludes the three Pi0.5 optional source families. Python compilation and diff whitespace checks pass. Proceeding with sequential squash merges.

PR #207, opened 2026-09-17
pi05: cache prompt embeddings and batch-width contexts

LiangSu8899:
## Scope

Split out of #202. Base: `feat/pi05-nvfp4-prefix`. This PR owns the related prompt-embedding and batch-width context caches.

- Reuse exact prompt/state embeddings and retain separate contexts for configured batch widths.
- Preserve per-slot valid prompt lengths when saving/restoring contexts.
- Keep mutation inside the frontend lifecycle lock; reload invalidates prompt embeddings and refreshes live pipelines.
- No kernels, build options or default precision changes.

## Validation

`pytest -q tests/test_pi05_prompt_cache.py tests/test_pi05_batched_widths.py`: **6 passed, 0 skipped** on this intermediate branch, including a regression for the production batch-width switch lifecycle guard.

Python compilation and `git diff --check` pass. Performance comparisons must label whether observations revisit already-cached prompt/state combinations.


LiangSu8899:
Review follow-up complete: state-bearing prompts now bypass the embedding cache, while task-only prompts retain the bounded LRU. A production-helper regression submits 600 distinct states, verifies fresh embeddings and no cache growth, and confirms that a task-only cache entry still hits. Documentation now describes this policy.

The complete CUDA 13 / RTX 5090 stacked suite passed 66 tests with no skips, including prompt rotation, reload invalidation, width switching and mixed-slot numerics. #205 and #206 have merged. This branch was rebased onto the new main and retargeted to main; the tree is identical to the tested cache stage and the diff contains only cache/width functionality.

PR #208, opened 2026-09-17
pi05: opt-in SDE denoising with isolated sampler kernels

LiangSu8899:
## Scope

Split out of #202. Base: `feat/pi05-prompt-width-cache`. Adds only the explicit stochastic denoising sampler.

- `FLASHRT_ENABLE_PI05_SDE=OFF` at build time and `sde=False` at runtime.
- Source, header and `pi05_sde_residual_add` binding share the same Pi0.5 gate, independently of Qwen.
- The skinny fused SDE declaration, implementation and binding require both skinny and SDE gates; SDE OFF exposes neither SDE binding even with skinny ON.
- Zero sigma matches the ODE path; deterministic noise/trace and per-slot batched sampling are covered.

## Validation

Final combined contract/kernel/frontend suite: **49 passed, 0 skipped** across 13 test files. Includes five SDE cases (BF16, cuBLASLt FP8, skinny FP8, trace/reproducibility and batched slots), and the production batch-width switch lifecycle guard.

Rebuilt SM120 core with all three options ON: core + FA2 import and all required symbols pass. Rebuilt with skinny/NVFP4 ON and SDE OFF: both SDE bindings and the fused SDE dynamic symbol are absent; core + FA2 import pass.

The final source was also rebuilt inside this container with all Pi0.5 feature gates OFF for SM87/89/110/120/121. All five core builds/links/imports pass; generated source lists and exported bindings exclude the new Pi0.5 kernels. The existing FA2 module imports alongside each core. These are cross-compilation/import checks, not numerical execution on five GPU types or five separately rebuilt FA2 modules.

Python compilation and `git diff --check` pass. The repository's existing asymmetric CFG precision failure is not hidden or relaxed; see the batch/prefix PR.


LiangSu8899:
Post-rebase validation: #201, #203, #204 and #202 have now been squash-merged in that order. The remaining #205 -> #206 -> #207 -> #208 stack has been rebased and pushed; none of these four downstream PRs has been merged.

RTX 5090 / consistent CUDA 13 container: the all-enabled core extension rebuilt and linked successfully. The combined downstream suite returned **50 passed, 1 failed**. The new mixed-slot B=4 versus independent B=1 test failed with per-slot cosine values [0.99976594, 0.99973133, 0.99864898, 0.07486279]. A fresh-process isolated rerun reproduced exactly the same values. Calibration scale equality and bit-exact slot permutation checks passed. This is NOT an acceptable numerical result and the gate has not been relaxed.

As a control, the merged-main-equivalent tree passed the same preceding seeded-noise, pooling, tokenizer and batched test sequence: **15 passed**. The downstream failure has not yet been bisected to an individual PR or feature. Keep the remaining stack unmerged until it is diagnosed and fixed. No performance claim is made from this validation.

LiangSu8899:
## Mixed-slot numerical mismatch: root cause and fix

The failure existed in the original, unsplit #202. It was not introduced by splitting or rebasing the PR stack.

### Controlled reproduction

RTX 5090, CUDA 13 container, identical checkpoint, observations, initial noise and calibration seed. The historical Python trees were tested against the same compatible native extension with the decoder explicitly pinned to cuBLASLt. This isolates the Python regression; it is not a claim that every historical native build/profile was rebuilt or tested.

| Tree | Minimum B=4 versus independent B=1 action cosine |
| --- | --- |
| Original #201, `a7099689` | 0.9937151057 |
| Original #202, `dcc23522` | 0.0748627863 |
| Split #205 before fix, `52d3094a` | 0.0748627863 |
| Split #205 after fix, `610450eb` | 0.9937151057 |

In each case the four slots have different synthetic observations and initial noise. Calibration scales match between B=1 and B=4, and permuting the slots preserves their outputs bit-for-bit. After the fix, **both the four B=1 outputs and the four B=4 outputs are bit-identical to the original #201 outputs**.

### Root cause

Original commit `56a996e7` replaced the FP8 weight quantizer's Python-double scale calculation and scalar division with FP32 device scale arithmetic and tensor division. That changes rounding at FP8 midpoint boundaries. Converted BF16 checkpoint tensors themselves are equal; the quantization change is sufficient to cause the observed downstream divergence.

The fix retains device-only execution: compute the scalar scale in FP64, cast to FP32, and multiply by its reciprocal to preserve the original scalar-division contract. There is no `.item()`, CPU transfer, new inference synchronization, kernel change, or relaxed numerical threshold.

### Verification

- Original #201 related tests: **11 passed**.
- Original #202 related tests on the controlled cuBLASLt route: **19 passed**. Those older tests do not exercise the newly added mixed-slot regression, which independently reproduces the failure above.
- Added **9 tests** for exact legacy quantized bytes/scales, zero and small scales, noncontiguous input, and CUDA Graph capture/replay with changed input.
- Fixed #205 quantization plus mixed-slot tests: **10 passed**.
- Full updated #205-#208 stack, including seeded noise, pooling, tokenization, mixed batching, skinny kernels, build contracts, reload lifecycle, NVFP4, prompt/width caches, SDE and the new quantizer tests: **60 passed, 0 skipped**.
- Changed Python compilation and `git diff --check`: passed.

The correction has been propagated through #205-#208. This supersedes the earlier unresolved 50-pass/1-fail receipt. No additional PR has been merged. These results establish compatibility for the tested inputs, not universal BF16/FP8 equivalence or an official-model accuracy claim.

LiangSu8899:
Review follow-up complete: SDE sources now live under csrc/kernels/pi05/pi05_sde_step.{cu,cuh}; CMake/includes and ownership tests match. The SDE option prints ENABLED/DISABLED status. Sigma schedules now reject NaN, positive/negative infinity and negative values before device staging, with four regression cases. Constructor keyword indentation is corrected.

RTX 5090 / consistent CUDA 13 container: all-enabled core rebuild/link succeeded; core and FA2 imports passed; NVFP4, SDE, skinny and fused-SDE symbols are present. All-OFF configure excludes the new Pi0.5 source families. Complete regression: **66 passed, 0 skipped**. Changed Python compileall and diff whitespace checks passed.

#205 -> #206 -> #207 have been squash-merged sequentially. This branch is rebased onto the resulting main and retargeted to main; its tree is byte-for-byte identical to the tested final stack and its remaining diff is the SDE increment. Proceeding with the final squash merge.

LiangSu8899:
Performance clarification: the earlier ~60 ms baseline included a timed set_prompt(prompt, state=...) on every frame. It must not be labeled default steady-state infer() latency or used to claim a default 60-to-12 ms speedup.

Remeasured merged main 337309e2 on RTX 5090 / CUDA 13: default FP8 + cuBLASLt, B=1, two camera views, 10 denoising steps, prompt initialized once (no per-frame state-in-prompt update), 30 warmups and 200 synchronized wall-clock measurements. Complete observation/image preprocessing -> vision -> prefix -> all denoising steps -> final action return: P50 19.610 ms, P95 19.839 ms, all outputs finite. This is not a partial-graph measurement or a per-slot amortized number. The historical ~12 ms opt-in result used a different cached state-prompt workload and is not a measurement of current default main.

PR #209, opened 2026-09-19
pi05: remove incorrect pre-unnormalize clip(-1, 1)

limezc:
## What

`unnormalize_actions()` (`flash_rt/core/utils/actions.py`) clips the raw model output to `[-1, 1]` *before* applying the q01/q99 quantile de-normalization:

```python
clipped = np.clip(actions, -1.0, 1.0)
unnorm = clipped.copy()
unnorm[..., :dim] = (clipped[..., :dim] + 1.0) / 2.0 * (q99[:dim] - q01[:dim] + 1e-6) + q01[:dim]
```

openpi's reference implementation (`openpi.transforms.Unnormalize._unnormalize_quantile`, which trained the checkpoints this frontend serves) does **not** clip before de-normalizing:

```python
return (x + 1) / 2 * (q99 - q01) + q01
```

## Why this is a real bug

For any action dimension whose raw model output exceeds `+/-1` (common for pi0.5 policies -- the denoised output is not guaranteed to stay inside the training quantile range), the clip silently saturates that dimension to the q99 (or q01) boundary *regardless of the model's actual output*. That produces a materially wrong, constant action value for that dimension instead of the value the model actually predicted -- a silent correctness bug, not a crash, so it is easy to miss without an independent reference to diff against.

## How it was found

Comparing FlashRT's FP8-quantized pi0.5 inference against openpi end to end on real LIBERO observations (same checkpoint, same physical GPU, same chunk size, real-camera-frame FP8 calibration) turned up one action dimension pinned to a constant value across every step, which traced back to this clip. Removing it and re-running raised the action-output cosine similarity between the two implementations from a degenerate constant-valued output to **0.9976**.

## Fix

Remove the pre-unnormalize clip; `unnormalize_actions` now matches `openpi.transforms.Unnormalize._unnormalize_quantile` exactly. No public API change -- same function signature, same call sites (both the single-sample and the batched `infer_batch` paths already funnel through this one function).

## Testing

- Direct check on an RTX 4090 that `unnormalize_actions` no longer clips out-of-range inputs and matches the openpi formula bit-for-bit for a synthetic action vector with entries outside `[-1, 1]`.
- `tests/test_pi05_cfg_batched_inference.py` on the same 4090 with a real pi0.5/LIBERO checkpoint: the non-subprocess tests (`test_rl_only_uses_serial_cfg_pipeline`, `test_rl_and_batched_builds_cfg_batched_pipeline`, `test_batched_cfg_inference_returns_finite_actions`) pass. The remaining cases in that file spawn several full-model-loading subprocesses per test and hit CUDA OOM on this single 24GB card independent of this change (pre-existing memory pressure in that test's design, not something this diff touches -- flagging this in case it is useful, but it is unrelated to the one-line fix here.
BODY
)


LiangSu8899:
Thx for fixing!! Looks like this came from some of my early simulation/eval adapter logic and somehow made its way into the shared utilities.

